PR #1, opened 2015-06-05
make gc not eat the tray

krl:
This seems to be needed both for mainWindow and the tray.


krl:
also, crashreporter was sending crash data somewhere, removed it.

jbenet:
lgtm

PR #13, opened 2015-06-21
Html menu

krl:
add init helper, make menu start and stop daemon, display stats, console and browser link.

firststart

click-tray

initialize

initialized

stopped

starting

running

console

browser


krl:
This depends upon the changes to node-ipfsd-ctl being merged.

whyrusleeping:
so pretty! :)

jbenet:
@krl solid progress! congrats on getting this working. Looking very good!

I've some feedback around the design, but that's a separate story than getting this stuff working in the first place -- so maybe we can merge this and finding the right design can be done in a future PR (and should probably iterate on the design in mockups a few times till we find something we really like + has great UX).

deploy.sh
```diff
@@ -0,0 +1,5 @@
+standard index.js || exit
+rm -R pkg
+sudo rm -R /home/mumin/Desktop/x64
+npm run package
+sudo cp -R pkg/0.1.0/linux/x64 /home/mumin/Desktop
```
jbenet: probably dont want to commit this `deploy.sh` file

html/help.html
```diff
@@ -0,0 +1,30 @@
+<html>
+		<head>
```
jbenet: can we indent html with either 2 spaces, or only 1 tab? 2 tabs indents crazy far.

jbenet: (and, indentation in html is not a big deal, just want to be able to read it on github nicely)

index.js
```diff
@@ -1,71 +1,186 @@
...
+      ipc.emit('node-status', 'stopped')
+    }
+  })
+
+  ipc.on('start-daemon', function () {
+    ipc.emit('node-status', 'starting')
+    node.startDaemon(function (err, ipfs) {
+      if (err) throw err
+      ipc.emit('node-status', 'running')
+      poll = setInterval(function () { pollStats(ipfs) }, 500)
```
jbenet: is this to refresh the menu? every 500ms may be intensive, particularly when minimized.

wonder if we can make the polling only happen while the menu is up?

index.js
```diff
@@ -1,71 +1,186 @@
...
   var parts = multiaddr(apiAddr).nodeAddress()
   var url = 'http://' + parts.address + ':' + parts.port + WEBUIPATH
   return url
 }
 
-var openWindow = function (ipfs) {
-  if (!mainWindow) {
-    ipfs.config.get('Addresses.API', function (err, res) {
+var openConsole = function () {
+  if (Ipfs) {
```
jbenet: should probably return an error if not, so maybe this func should take an optional callback

index.js
```diff
@@ -1,71 +1,186 @@
...
-        click: function () { openWindow(ipfs) }
-      },
-      { type: 'separator' },
-      { label: 'Quit',
-        click: function () { process.exit(0) }}
-    ]))
-  })
-})
+var openBrowser = function () {
+  if (Ipfs) {
```
jbenet: same here

js/initialize.jsx
```diff
@@ -0,0 +1,147 @@
...
+    })
+  },
+
+  render: function () {
+    var t = this
+    var initializing = (t.state.initState === 'initializing')
+    var initialized = (t.state.initState === 'initialized')
+    var uninitialized = (t.state.initState === 'uninitialized')
+
+    var description = 'Before starting, ipfs needs to generate a keypair, and create a directory to live in.'
```
jbenet: all of this should probably live on html pages/templates that we can easily tweak. we'll want to play with all this stuff quite a bit as time goes on.

jbenet: How about: 

> Before starting, IPFS Node needs to generate a keypair, and create a directory for data to live in.

jbenet: Separately, i think we want to add two checkboxes to this initial "setup dialog":

- [x] Add IPFS Node to Login Items[0]. This opens the IPFS menubar item. If it is switched to "On", it will also start your IPFS Node.[1]

[0] this is platform specific. may need different language per platform.
[1] there's got to be a better, simpler way to say this. the tricky part is it must capture (a) the on/off switch, (b) "IPFS Node" app remembering the state user left it in (on/off), (c) that it will start the IPFS node iff the node was left "On". (all without being too verbose. design is hard :) ).

js/initialize.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+    var success = 'Your ipfs node was setup successfully, you can now close this window and start using the ipfs tray toolbar.'
```
jbenet: > You can now close this window, and start using the ipfs tray toolbar.

this is platform specific and somewhat unclear.


- we say:
  > Your IPFS Node was setup successfully! You can control your node, and access the Web Console, from this new toolbar menu:
  >
  > < show-platform-specific-screenshot-of-the-menu >

- The dialog has "Close" focused and clearly visible.
- We pop open the Menu (and highlight the tray icon or whatever), so the user can see it.
- Make the node be already running, so the user only has to do something if they want to turn it off (alternatively, we could start with the node _off_ to _train_ the user to turn it on and off, but this may not be necessary).

js/initialize.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+    var error = (uninitialized && t.state.error)
+      ? (<pre className='alert alert-danger nomargin top-buffer'>
+          {t.state.error}
+         </pre>) : null
+
+    var loader = initializing ? (
+      <img className='loaderanimation pull-right' src='../img/loading.gif'/>
```
jbenet: nice!

pkg.js
```diff
@@ -8,9 +8,9 @@ var path = require('path')
 var sh = require('shelljs')
 
 var appVersion = pkgjson.version
-var appName = pkgjson.name
+var appName = 'IPFS Node'
 var electronPackager = './node_modules/.bin/electron-packager'
-var electronVersion = '0.26.0'
+var electronVersion = pkgjson.devDependencies['electron-prebuilt']
```
jbenet: what's the implication of this? it locks to whatever we have in package.json?

jbenet:
Comments above o/

jbenet:
thanks for the screenshots. very useful for people following along :+1: :+1:

pkg.js
krl: yep

js/initialize.jsx
krl: #15

jbenet:
@krl is the `+33,492 js/initialize.js` thing supposed to be committed?

jbenet:
@krl what extra review is needed here? i dont have time to read all the changes carefully atm. Im fine to merge if you think it's solid. happy to look at something closer, or read the whole thing later on.

if others (@travisperson or @bcomnes) want to CR, much appreciated

index.js
```diff
@@ -1,71 +1,202 @@
...
+var Ipfs
+
+var error = function (err) {
+  mainWindow = new BrowserWindow({icon: LOGO,
+                                  'auto-hide-menu-bar': true,
+                                  width: 800,
+                                  height: 600})
+  mainWindow.loadUrl('file://' + __dirname + '/html/help.html')
   mainWindow.webContents.on('did-finish-load', function () {
+    console.log('in loaded')
```
travisperson: We should probably add a debugger module. [debug](https://www.npmjs.com/package/debug) is always my go-to, but I have no real preference.

(though this line may not even be needed for debugging)

js/initialize.jsx
```diff
@@ -0,0 +1,147 @@
...
+      })
+    })
+
+    ipc.on('initialization-complete', function () {
+      t.setState({initState: 'initialized'})
+    })
+  },
+
+  setKeySize: function (size) {
+    var t = this;
```
travisperson: Semicolon, also if we are writing this with standard, `this` should be assigned to `var self` correct?

jbenet: CI should probably run standard and fail the build? pre-commit hooks?

travisperson: Possiblly, but I think it may cause us to run into problems. Sometimes things have to be ignored. I think it might be better to just handle them in PRs. Though if there is a way to do warnings or something that would be neat.

The issue I see (and might be why this slipped in) is on line 105 of this file there is an HTML entitie. This may of failed standard and required a no veriffy to commit it.

There may also be a way around it with some variable trick.

krl: t -> self, agreed. There  was no complaints on this though, just copied the style from webui.

jbenet:
@travisperson comments addressed? RFM?

travisperson:
:+1:

krl:
Yes this is ready, will add an issue for some of the comments that got eaten by github

PR #18, opened 2015-06-29
make sure `npm start` runs `./build.sh`

jbenet:
Otherwise assets will not build correctly. Maybe there should be
a check somewhere, as it took me a while to figure out what was wrong


jbenet:
cc @krl

PR #27, opened 2015-08-06
restructure codebase

diasdavid:
This is a big PR that essentially makes the codebase more organized an nice. It's cleared what happens in 'core side' and 'view side', what is the `app` code, what is the `build` and what is `dist`.

Also fixes some things that were not handled correctly + removes dead code.

I believe that there might be some space to improve the structure of each react component (privilege reusability), but I would prefer to tackle that as the rest of the features get developed and bugs fixed, using this as the code structure from now on, therefore, PR'ing :)


jbenet:
LGTM!  :+1:

jbenet:
Muuuuuuuch nicer

diasdavid:
woooot ! :D \o/ Thank you :)

PR #30, opened 2015-08-11
drag and drop feature

diasdavid:
- [x] drop to upload a file
- [x] copy gateway url to clipboard
- [x] show latest 5 files uploaded
- [ ] make it pretty


diasdavid:
Use https://github.com/ipfs/electron-app/pull/32

PR #157, opened 2016-01-06
Change example clone to be HTTPS based

Kubuxu:
You need to have SSH key that is assigned to GH account to be able to clone using SSH.


dignifiedquire:
Thanks

PR #198, opened 2016-02-22
Better error handling and fix file uploading.

### LINKED ISSUE #153: Window drag and drop file upload is broken
Probably the same issue as https://github.com/sarink/react-file-drop/issues/5

dignifiedquire:
Closes #153


app/controls/drag-drop.js
```diff
@@ -1,10 +1,7 @@
-import API from 'ipfs-api'
 import notifier from 'node-notifier'
 import {join} from 'path'
 import {getIPFS, logger} from './../init'
```
diasdavid: this works in Node.js due to transpilation, correct? Do we want to use two module systems in the same file? I'ld rather prefer to keep using CJS on Node.js side and ES6 for browser/ui side.

app/controls/drag-drop.js
```diff
@@ -33,29 +30,35 @@ function notifyError (message) {
...
-    if (err || !res) {
-      logger.error(err)
-      notifyError(err || 'Failed to upload files')
-    }
 
-    logger.info('Uploading files', {files})
+  ipfs.add(files)
+    .then((res) => {
+      if (!res) {
+        notifyError('Failed to upload files')
```
diasdavid: missing a `return` here

app/controls/drag-drop.js
dignifiedquire: The whole code base already uses es6 modules as everything gets transpiled anyway. If they are different repos we can have that differentiation, but in this case I'd rather have consistency in the repo. 

(Also this is not changed in this PR)

app/controls/drag-drop.js
```diff
@@ -33,29 +30,35 @@ function notifyError (message) {
...
+
+      logger.info('Uploading files', {files})
 
-    res.forEach((file) => {
-      const url = `https://ipfs.io/ipfs/${file.Hash}`
-      clipboard.writeText(url)
-      filesUploaded.push(file)
+      res.forEach((file) => {
+        const url = `https://ipfs.io/ipfs/${file.Hash}`
+        clipboard.writeText(url)
```
diasdavid: I guess this works because in a context of adding a directory, the 'last file' is actually the root hash of the whole dir. This doesn't work as nicely when a group of files is added, they needed to be wrapped in a dir object (like `ipfs add path -w`)

dignifiedquire: This doesn't do directory uploads, as browsers still don't allow us to do it :cry:

Does `-w` work on the api?

app/controls/drag-drop.js
diasdavid: noted, why have `const clipboard = require('clipboard')` (line 8) then? :+1: for consistency

app/controls/open-browser.js
```diff
@@ -1,23 +1,24 @@
 import ipc from 'electron-safe-ipc/host'
 import {apiAddrToUrl} from './utils'
-import {getIPFS} from './../init'
+import {logger, getIPFS} from './../init'
```
diasdavid: can we get the `logger` coming from a `config.js` file, along with other default values?

app/init.js
```diff
@@ -21,36 +23,42 @@ if (config.isProduction) {
...
+    .then(next)
+
+  // ipfs.id((err, peer) => {
+  //  if (err) throw err
+  // lookupPretty(ipfs, peer.Addresses, (err, location) => {
+  //   if (err) throw err
+  //   statsCache.location = location && location.formatted
+  //   ipc.send('stats', statsCache)
+  // })
+  // })
```
diasdavid: do we still need this code?

app/controls/drag-drop.js
dignifiedquire: fixed :)

app/init.js
dignifiedquire: yes, as soon as geoipfs is fixed

app/controls/open-browser.js
dignifiedquire: No because it is not coming from the config but from the main index file that starts everything up

app/controls/drag-drop.js
dignifiedquire: hmmm no idea why that's there oO

dignifiedquire: fixed

app/controls/drag-drop.js
diasdavid: > This doesn't do directory uploads, as browsers still don't allow us to do it 

But we can drag and drop several files to the tray icon :) 

> Does -w work on the api? 

It should :)

app/init.js
diasdavid: got it, mind having a TODO and a link to the issue saying that?

dignifiedquire: done

diasdavid:
LGTM.

There is no tests still, makes contributions really hard to do.

dignifiedquire:
> There is no tests still, makes contributions really hard to do.

well you could write some :P

PR #283, opened 2016-06-04
feat: update peer stats only when profile window is open

### LINKED ISSUE #60: Update stats on window show
At the moment depending on the timeout position there is a delay between opening the profile screen and the stats updating. We should listen for the window `show` event and update the stats at that point.

swang:
Please let me know if my assumptions about things were correct (like shouldPoll being unnecessary if we're triggering from the 'show' event.)

---

Polling for the number of peers is now only updated when the user opens
the profile window rather than having a timer constantly updating in the
background.

Got rid of `shouldPoll` because it seems unnecessary with the change in
actually shown.

Added a new event handler to trigger the polling when the window is
how timers are set. The application also doesn't immediately run
`pollStats` when loading the ipfs daemon.

Closes #60.

### DIFF
--- a/app/init.js
+++ b/app/init.js
@@ -25,20 +25,16 @@ let IPFS
 let ipc
 let mb
 let logger
-let shouldPoll = false
 const statsCache = {}
 
 function pollStats (ipfs) {
   const next = () => {
     setTimeout(() => {
-      pollStats(ipfs)
+      if (mb.window.isVisible()) {
+        pollStats(ipfs)
+      }
     }, 1000)
   }
-
-  if (!shouldPoll || !mb.window || !mb.window.isVisible()) {
-    return next()
-  }
-
   ipfs.swarm.peers()
     .then((res) => {
       statsCache.peers = res.Strings.length
@@ -79,9 +75,6 @@ function onStartDaemon (node) {
     if (err) throw err
     ipc.send('node-status', 'running')
 
-    shouldPoll = true
-    pollStats(ipfsNode)
-
     ipfsNode.version()
       .then((res) => {
         ipc.send('version', res)
@@ -98,11 +91,9 @@ function onStopDaemon (node, done = () => {}) {
   logger.info('Stopping daemon')
 
   ipc.send('node-status', 'stopping')
-  if (shouldPoll) {
-    delete statsCache.peers
-    ipc.send('stats', statsCache)
-    shouldPoll = false
-  }
+
+  delete statsCache.peers
+  ipc.send('stats', statsCache)
 
   node.stopDaemon((err) => {
     if (err) return logger.error(err.stack)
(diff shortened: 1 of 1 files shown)


swang:
i know there is a conflict here from the recent merges but should i bother fixing this? this pr has sat here for ~4 months without nary a comment. is it okay? does it need changes?

i understand people are busy so i'm not pushing for an immediate merge. just want to know if i should spend time to fix the conflict issue.

dignifiedquire:
So sorry about the lack of feedback, let me review it now

app/init.js
```diff
@@ -25,20 +25,16 @@ let IPFS
...
 let mb
 let logger
-let shouldPoll = false
 const statsCache = {}
 
 function pollStats (ipfs) {
   const next = () => {
     setTimeout(() => {
-      pollStats(ipfs)
+      if (mb.window.isVisible()) {
```
dignifiedquire: please check for `mb.window` I have seen it being undefined in edge cases

dignifiedquire:
Looks pretty good to me, left some notes. If you could rebase & fix those I can merge it in

swang:
hey so i tried running the latest master and was getting some errors. i tried reinstalling and now i just get a blue screen with a loading circle in the middle. are you seeing this as well?

swang:
so i drilled down and found out that in `ipfsd-ctl` it makes subcalls to `go-ipfs` and go-ipfs is complaining about a repo/package that is not up to date. this message gets drowned out in startDaemon and thus never propagates up to the surface.

this is what i see when i try to run `./ipfs daemon`

```
Initializing daemon...
Adjusting current ulimit to 1024...
Successfully raised file descriptor limit to 1024.
Found outdated fs-repo, migrations need to be run.
Run migrations now? [y/N]
```

VictorBjelkholm:
Try clearing (removing) your `$IPFS_PATH` folder or migrate it manually
with https://github.com/ipfs/fs-repo-migrations

*Victor Bjelkholm*
Software Engineer
Protocol Labs, Inc.
http://ipn.io | https://ipfs.io

diasdavid:
@hacdias are you taking this into account with the Redesign PR #516 ?

hacdias:
No, but I will. @diasdavid

diasdavid:
This was already integrated, see comment here https://github.com/ipfs-shipyard/station/pull/516#issuecomment-348679352. Thanks for the PR either way @swang :)

swang:
The issue wasn't that polling didn't activate when you opened the window. The issue was that it continued to poll once you closed the window.

You can look at the OP of this PR and the diff in the code and also look at the comments in #60.

hacdias:
If you look at [this line](https://github.com/ipfs-shipyard/station/pull/283/files#diff-c0678b66430c9e98d807b2f04b038930L38) it will not poll if the window is closed. Yes, it will call the function again, but the API of IPFS never gets called.

swang:
I guess I must have been confused then. I vaguely remember checking if it made the api calls beforehand to see if they fired after the window closed.

PR #509, opened 2017-06-03
Redesign

dimoFeeld:
I've quickly redesigned the Station app. It's far from final and the file is dirty but if you guys are happy let me know. I can see the project becoming bigger as a serious dropbox alternative as longs we we manage to make it user friendly enough.

Feedback is welcome


lidel:
I'd love to have a chance to glance at it, but it is in proprietary Sketch format.
Could you [export it to PDF](https://www.sketchapp.com/learn/documentation/exporting/file-formats/) please, so it can be viewed by non-Mac users?


(I am truly interested in this, as we want to unify UX across "official" apps and use similar design language in browser addon ([ipfs/ipfs-companion](https://github.com/ipfs/ipfs-companion)))

diasdavid:
This is a really cool contribution @dimoFeeld ! Thank you :)

Will you have the time to keep iterating on the design? Would love to see a new design come to life :) Agree that UX is crucial!

Here is my first feedback:
image

dimoFeeld:
@lidel I just exported it [as a pdf](https://github.com/dimoFeeld/station/blob/df27ad051795c3cd99de9b83c6030e731d16962e/design%20resources/Previews/IPFS%20Station.pdf). What do you think? It's at early stage but my concept is to package the whole thing as super duper user friendly app. I will keep expanding on it, maybe even work on some sort of a brand for it.

I think it will be great if we manage to package everything under a single domain where we can let people download their browser plugins and desktop and mobile apps.

dimoFeeld:
@diasdavid Thank you so much for the feedback. It makes total sense. I am working on it as we speak. But keep in mind that my goal is to attract the mainstream users to the platform. So I would need your assistance so I can make sure that the power user features are secondary.

lidel:
@dimoFeeld thanks, [PDF renders directly in github interface](https://github.com/dimoFeeld/station/blob/df27ad051795c3cd99de9b83c6030e731d16962e/design%20resources/Previews/IPFS%20Station.pdf)  👍 

Few thoughts:

- I like minimalist approach, but just like @diasdavid  noted, bottom icons should be self-explanatory: 
  we should not force user to discover meaning via on-hover tooltip.  Perhaps something like https://thenounproject.com could be a good source of meaningful icons?
- Non-macOS use cases such as browser extension will not be able to support support translucent popups. This means if you don't provide backgrop image, the background will be just a solid dark grey/black. Nothing wrong with that, but the "solid one" will have more use cases: it means you need to prepare two mockups for the design: one with translucent background, and one with solid one.

dimoFeeld:
@lidel @diasdavid I've added a solid colour background as an option in the designs. Icons are much better now. And you can also see the peer's IDs on the map.

diasdavid:
Hi @dimoFeeld, @hacdias is continuing and improving on your initial work at #516 . I'm closing this PR but you are welcome to join #516 too :)

PR #511, opened 2017-11-22
[WIP] Update Dependencies and make this work

hacdias:
- [x] Update Dependencies
- [x] Fix Small bugs
- [x] Fix Welcome Screen
- [x] Move from radium to react inline styling which is exactly the same (https://reactjs.org/docs/dom-elements.html#style)
- [x] Stateless views are now functional


hacdias:
GREEN! @diasdavid

package.json
```diff
@@ -4,50 +4,52 @@
...
     "react-dnd": "^2.1.2",
     "react-dnd-html5-backend": "^2.1.2",
-    "react-dom": "^15.1.0",
-    "react-widgets": "^3.2.0",
+    "react-dom": "^16.1.1",
+    "react-widgets": "^4.1.1",
     "winston": "^2.1.0"
   },
   "devDependencies": {
     "babel-core": "^6.3.21",
```
diasdavid: While you are at it, can you add `aegir` as a devDep and use its `aegir lint` to make sure the linting rules are the same as the remaining modules?

package.json
```diff
@@ -4,50 +4,52 @@
   "description": "IPFS Native Application",
```
diasdavid: Let's take off the `1.0.0-alpha.1`, replace by `0.1.0` (that is a old pratice)

src/init.js
```diff
@@ -43,17 +43,23 @@ function pollStats (ipfs) {
     })
     .then(next)
 
+  // TODO: solve?
   ipfs.id()
     .then((peer) => {
+      console.log(peer)
```
diasdavid: Unnecessary console.log

package.json
hacdias: Should I add it to webui too?

package.json
hacdias: Done!

package.json
hacdias: Done done 😄

hacdias:
@diasdavid changed made!

package.json
```diff
@@ -53,8 +56,8 @@
   },
   "scripts": {
     "start": "electron src/index.js",
-    "test": "aegir lint",
-    "lint": "aegir lint",
+    "test": "eslint .",
```
diasdavid: We really need some tests here :)

diasdavid:
@hacdias is drag and drop working?

hacdias:
@diasdavid yup, it's working. It doesn't send any message telling so tho.

hacdias:
I see we're using notifier to send notifications. Windows 10 API has changed in the last update and I think that's the reason why I'm not receiving any notification. Source: https://github.com/mikaelbr/node-notifier/issues/208.

PR #512, opened 2017-11-25
[WIP] Packaging using Electron Forge

hacdias:
https://electronforge.io/

- [x] Review, clean and comment the code.
- [x] Make function to get IPFS logo.
- [x] Pretty Logger


hacdias:
@diasdavid personally I don't enjoy the 'index.js' very much (previous init.js). I feel it too heavy, I don't know. I think I should clean it.

One question: what is supposed to be the meaning of 'controls' directory?

hacdias:
@diasdavid according to Electron Forge page,

>Makes platform specific distributables of your Electron application. Currently you can only generate distributables for your current platform, this means for instance you can't generate Windows installers from a linux machine.
>
> We reccomend running this command on CI (normally Travis CI and Appveyor) so that you can generate distributables for all 3 platforms.

So, it can **package** for every platform on one machine but it can't make the **distributables**. Windows distributables can only be made on Windows so I'd suggest to do it on AppVeyor and darwin/linux can all be made on Travis.

Right now, the only thing left to configure is the automatic "packaging" when releasing a tag and then the upload to some place. I think we should do that when we think it's time to start packaging.

diasdavid:
@VictorBjelkholm is our CI master. @VictorBjelkholm can you make this distributables happen?

VictorBjelkholm:
Yeah, looking into this right now, should be quick. Need to know what commands to be run and where the distributables ends up so we can archive them, then I'll add a custom pipeline for station

diasdavid:
@hacdias shall we merge this PR and continue the publishing distributables as an issue?

hacdias:
Yes, I think that would be better so I can start thr redesign in a new
Branch with the latest code. :)

PR #513, opened 2017-11-28
[WIP] Redesign

hacdias:
Based on #509 

- [x] Copy To Clipboard button that shows on hover
- [x] Show user location instead of address
- [ ] Files
- [ ] Settings
- [ ] Peers Search
- [ ] Peers Additional Settings
- [ ] Glowing Icon Top Right
- [ ] Add files to extensions list
- [ ] Integrate Welcome Screen on Menubar instead of different window 😄 
- [ ] Improve Welcome Screen with new design
- [ ] Remove useless font styles

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-* text eol=lf


diasdavid:
Past screenshots as you go :D

hacdias:
image

image

Here (^) are some screenshots of the latest commit on this branch. I'd like to get some feedback about some topics:

- Should the left panel (peers) be **always** expanded? If not, should it be like [this](https://github.com/ipfs-shipyard/station/pull/509/files#diff-db59a52c22f1e964b84200e6f949e1ac) or something else?
- There are going to be three different panels on the right side: Your Files, Your Node and Settings. Which one should be the "default"? How to do a smooth navigation between the three? I thought about removing the three dots in the bottom and:
    - When you're on the default, there are two different buttons: one for each of the other panels.
    - When you're not on the default, there is one button: an arrow to go back.
- Should I put a close button on the top right? (I vote yes) According to #509 that bit was reserved for actions like "add file" or "add folder". What do you think?

/cc @diasdavid @VictorBjelkholm @dimoFeeld

hacdias:
image

diasdavid:
Just gave it a test and I like it!

image

I know there are still work to do (for example, it doesn't show the files yet) so I'll avoid giving feedback there but here are a couple of points on the info menu:

image

image

hacdias:
Sorry, I had the protocol and agent versions swapped! Which one should I remove?

hacdias:
@diasdavid any feedback about this questions: https://github.com/ipfs-shipyard/station/pull/513#issuecomment-347612399

hacdias:
image

diasdavid:
> Should the left panel (peers) be always expanded? If not, should it be like this or something else?

Let's stick with it expanded now and reevaluate after some usage :)

> There are going to be three different panels on the right side: Your Files, Your Node and Settings. Which one should be the "default"? 

What about merging the "Your Node" and "Settings" pane into one and then using the left pane to switch between peers and files added?

> Should I put a close button on the top right? (I vote yes) According to #509 that bit was reserved for actions like "add file" or "add folder". What do you think?

Not sure. Show me how it looks! :)

Btw, check out https://github.com/ipfs/ipfs-companion/pull/320#issuecomment-347877262

Can we have a glowing animation of IPFS to demonstrate it's on/off state too??

hacdias:
> Can we have a glowing animation of IPFS to demonstrate it's on/off state too??

Yes, we can! But where? 😮 I'll try to find a good place to put it, but I'll be waiting for feedback 😄

diasdavid:
@hacdias perhaps you can reduce its size and keep it as corner element

hacdias:
image

image

image

package.json
```diff
@@ -12,7 +12,7 @@
     "ipfs-api": "^17.1.3",
     "ipfs-geoip": "^2.3.0",
     "ipfs-logo": "github:ipfs/logo",
-    "ipfsd-ctl": "^0.25.1",
+    "ipfsd-ctl": "^0.26.0",
```
diasdavid: Did it work flawlessly for you? I'm having a hard time here https://github.com/ipfs/js-ipfs-api/pull/638

hacdias: Haven't tested it yet. Probably later today, but can't now, unfortunately. 😞

hacdias: @diasdavid I'm not having issues (to test I deleted every IPFS binary from my computer, removed the ipfs repo and just run `station`).

diasdavid: You can add files still?

hacdias: Yes, I'm not having any issues.

PR #516, opened 2017-12-01
WIP - Redesign

### LINKED ISSUE #3: menu bar improvements
the menu bar should give some info on the node, like:
- say what ports it's on (swarm, gateway, api)
- stats (maybe under a sub-menu):
  - uptime
  - # of peers connected
  - bandwidth numbers

and maybe it should have some more options:
- "Open" should be "Open Web Console" or "Open Web UI" or "Open Admin Console" or something...
- "Open Web Console in Browser" (opens in a normal browser, just in case)
- "Configuration" (opens config in an editor)
- "Restart" (kills the node and reboots it)
- "Check for Update" (checks for update-- we need to figure this out)

### LINKED ISSUE #28: Add ~~`right click`~~ `alt click` for settings and quit
with -> https://github.com/atom/electron/blob/master/docs/api/tray.md#event-right-clicked

hacdias:
Please refer to #513 for the older conversation. Based on #509.

- [x] Copy To Clipboard button that shows on hover
- [x] Show user location instead of address
- [x] Remove useless font styles
- [x] Files
- [x] Integrate Welcome Screen on Menubar instead of different window 😄 
- [x] Close #28 (Context Menu for Settings and Exit)
- [x] Glowing Icon Top Right (https://github.com/ipfs/ipfs-companion/pull/320#issuecomment-347877262)
- [x] Close #283 (Poll stats only when Station is opened)
- [x] Streamline Buttons
- [x] Peers Search
- [x] Less focus on pub key
- [x] Add files to extensions list
- [x] Edit settings button on info
- [x] Improve Welcome Screen with new design
- [x] Button + and 📁 on Files
- [x] Close #3
- [x] Clean CSS
- [x] Clean JS
- [x] Better way to show errors in production
- [x] Update Readme


hacdias:
image

image

hacdias:
Context Menu:

image

hacdias:
image

Glowing icon! Perfectly beautiful. Somehow we will have to put it on WebUI!

diasdavid:
image

diasdavid:
@lidel, @hacdias just imported your glowing IPFS icon here too :D looks gooooood!!

diasdavid:
Also, @lidel pretty soon you can start recommending users to install an IPFS node through station so that they can use ipfs-companion with go-ipfs.

hacdias:
@diasdavid apparently, PollStats was already running *only* when the window was opened. I don't fully understand why the #283 was opened in first instance since the code in that time already polled only if the window was visible.

hacdias:
Ongoing welcome screen update:

image

The big logo at left is an heartbeat.

diasdavid:
image

hacdias:
image

image

diasdavid:
image

hacdias:
Hmm... I was just trying to let the sizes be consistent across the whole
UI. But we can change that, yes.

diasdavid:
@hacdias experiment :) You can change the sizes or increase that pane a bit :)

hacdias:
image

image

hacdias:
@diasdavid sometimes, when I hit the turn off button or close Station, IPFS daemon doesn't remove the 'lock' file and then it doesn't start the next time. Do you think it's related to IPFS itself or something with station?

hacdias:
@diasdavid I think that basically everything is done in this PR in terms of Features. Could you check #3 to see if there is any other feature worth implementing in there? If so, we could list them here and close #3.

hacdias:
@diasdavid one more thing: could you check if dropping files **and** folders to the tray works since it's a macOS only feature?

diasdavid:
> re: https://github.com/ipfs-shipyard/station/issues/3

Seems like the only thing missing is the bandwidth usage, but that sounds like it is a thing that should come from go-ipfs itself. Wanna create an issue with that feature request?

diasdavid:
@hacdias it is acting weird in Mac OS X.

The menu bar only shows 

image

With either left or right click, and it starts an IPFS daemon automatically (I checked the running processes).

hacdias:
@diasdavid probably the context menu overrides the event for left click on macOS. I don't know. But is there a need to have a context menu? Both actions are available on the right pane. What do you think? It is working on Windows and Mint.

diasdavid:
Ok, now cleaned my local repo and let the electron create a new repo, it is working but:

image

Also, the Window is not closing anymore. Clicking the menubar item now only shows the settings and exit options.

diasdavid:
Drag an drop of folders worked 🌟

hacdias:
Awesome! I think we should remove the context menu then. It turned blue before? :O

diasdavid:
Btw, Station needs one more feature. Adding to the files added, we need a screen to do Pinning!

diasdavid:
> It turned blue before? :O

AFAIR, yes :)

diasdavid:
>  But is there a need to have a context menu? Both actions are available on the right pane.

Yeah, get rid of it :)

hacdias:
I don't know anything about it being blue before and it has always been blue on Windows, so probably I can't help with that. But let's remove the context menu. Maybe it's what's causing this mess!

src/config.js
```diff
@@ -40,22 +51,10 @@ if (isDev) {
...
-}
-
 // Configuration for the MenuBar
 const menubar = {
   dir: __dirname,
-  width: 300,
+  width: 850,
   height: 400,
   index: `file://${__dirname}/views/menubar.html`,
   icon: (os.platform() === 'darwin') ? macOsMenuBar : getLogo(),
```
diasdavid: Can you change this icon on the fly? I believe that you can.

I see a dark gray icon because that is the default for MacOS, but before it was changing to the blue one when the node got started.

diasdavid: @hacdias this is the culprit for the dark gray IPFS logo on my machine.

hacdias: Did you try again after I removed the context menu? Cause I think it might be related!

diasdavid:
Now I can close and open the UI just fine:
image

src/config.js
diasdavid: Yes -- https://github.com/ipfs-shipyard/station/pull/516#issuecomment-349065402 --, still gray though

src/config.js
```diff
@@ -8,7 +8,18 @@ import {getLogo, macOsMenuBar} from './utils/logo'
 
 const isProduction = !isDev
 const currentURL = (name) => `file://${__dirname}/views/${name}.html`
-const ipfsPathFile = path.join(app.getPath('appData'), 'ipfs-electron-app-node-path')
+const ipfsAppData = (() => {
+  let p = path.join(app.getPath('appData'), 'ipfs-station')
```
diasdavid: This can be a `const`

src/config.js
```diff
@@ -8,7 +8,18 @@ import {getLogo, macOsMenuBar} from './utils/logo'
...
+    fs.mkdirSync(p)
+  }
+
+  return p
+})()
+
+const ipfsPathFile = path.join(ipfsAppData, 'app-node-path')
+const ipfsFileHistoryFile = path.join(ipfsAppData, 'file-history.json')
 
 const ipfsPath = (() => {
```
diasdavid: This function should use fs.existsSync instead of trying to catch for errors.

src/controls/upload-files.js
```diff
@@ -0,0 +1,27 @@
...
+      res.forEach((file) => {
+        const url = `https://ipfs.io/ipfs/${file.hash}`
+        clipboard.writeText(url)
+        logger.info('Uploaded file', {path: file.path})
+        appendFile(file.path, file.hash)
+      })
+    })
+    .catch((err) => {
+      logger.error(err)
+      // FAILED TO UPLOAD FILES
```
diasdavid: Unnecessary comment. Other reasons might happen too

src/controls/upload-files.js
```diff
@@ -0,0 +1,27 @@
+import {logger} from '../config'
+import {appendFile} from './../index'
+import {clipboard} from 'electron'
+
+export function uploadFiles (ipfs, event, files) {
+  ipfs()
+    .add(files, {recursive: true, w: files.length > 1})
+    .then((res) => {
+      if (!res) {
+        // FAILED TO UPLOAD FILES
```
diasdavid: Unnecessary check. If there is no res, then it means it has thrown an error

src/index.js
```diff
@@ -2,15 +2,13 @@ import menubar from 'menubar'
...
-import openSettings from './controls/open-settings'
 import openBrowser from './controls/open-browser'
-import dragDrop from './controls/drag-drop'
+import {uploadFiles} from './controls/upload-files'
 
 import {join} from 'path'
 import {lookupPretty} from 'ipfs-geoip'
 import config, {logger} from './config'
-import {dialog, BrowserWindow, ipcMain, app} from 'electron'
+import {dialog, ipcMain, Menu, shell, app} from 'electron'
```
diasdavid: My linter warned that `Menu` is not being used.

hacdias:
@diasdavid opened (https://github.com/ipfs/go-ipfs/issues/4456). Can we close #3?

src/index.js
```diff
@@ -40,21 +38,45 @@ let IPFS
...
 
   if (!shouldPoll || !mb.window || !mb.window.isVisible()) {
     return next()
   }
 
   ipfs.swarm.peers()
     .then((res) => {
-      statsCache.peers = res.length
+      res = res.sort((a, b) => {
+        return a.peer.toB58String() > b.peer.toB58String() ? 1 : -1
```
diasdavid: You can just do `return a.peer.toB58String() > b.peer.toB58String()`, it will already return true/false

src/js/components/view/file.js
```diff
@@ -0,0 +1,69 @@
...
+import PropTypes from 'prop-types'
+import {clipboard} from 'electron'
+import moment from 'moment'
+
+import TextButton from './text-button'
+import Icon from './icon'
+
+export default function File (props) {
+  let icon = 'file'
+  let extension = props.name.split('.').pop()
```
diasdavid: It might be better to use https://www.npmjs.com/package/file-extension

diasdavid: Also, use const for const

src/config.js
hacdias: Strange... I don't know what I might have changed related to that... And I can't really try that out 😮 Could you check that, please?

diasdavid:
@hacdias once you merge this PR, yes :)

src/config.js
diasdavid: @hacdias you can try it indeed. You also should experience on Windows

- Daemon off - Gray logo
- Daemon on - Blue logo

src/index.js
hacdias: IDK why I did that 😆

hacdias:
@diasdavid,right now, if there is an uncaught exception, the program will just close without any warning. THere's an app.log with the info, but don't you think we should say something to the user?

hacdias:
"Pin Files" moved to #518 and "Add Stats On Right Pane" to #523.

src/config.js
hacdias: done :)

diasdavid:
@hacdias agreed. There should be an infobox with that error and then try to graciously close the program.

src/controls/open-webui.js
```diff
@@ -0,0 +1,11 @@
+import {shell} from 'electron'
+import {apiAddrToUrl} from './utils'
+import {logger} from '../config'
+
+export default function openWebUI (ipfs, cb) {
+  ipfs().config.get('Addresses.API')
+    .then((res) => {
+      shell.openExternal(apiAddrToUrl(res))
+    })
```
diasdavid: nitpick, you can reduce this to just one line: `.then((res) => shell.openExternal(apiAddrToUrl(res)))`

src/controls/open-webui.js
```diff
@@ -0,0 +1,11 @@
+import {shell} from 'electron'
+import {apiAddrToUrl} from './utils'
+import {logger} from '../config'
+
+export default function openWebUI (ipfs, cb) {
```
diasdavid: Also, always please use `callback` as the top level callback of a function and `cb` for internal ones (makes them easier to distinguish)

src/config.js
hacdias: Done on 232ba2d

diasdavid:
Fantastic work @hacdias !!! 



This makes Station so much better 👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽

Issue #517, opened 2017-12-04
Trouble building on macOS


machawk1:
macOS 10.13.1
node v8.9.1
npm 5.5.1
station aa8b8630fc2e688693176c4ae06d5e3db13d9e77

Clone the repo, then cd'd into it, per the README.

<details>
<summary>npm install finished with some caveats</summary>
<pre>
> macos-alias@0.2.11 install /Users/machawk1/Downloads/station/node_modules/macos-alias
> node-gyp rebuild

gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
No receipt for 'com.apple.pkg.CLTools_Executables' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLILeo' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLI' found at '/'.

gyp: No Xcode or CLT version detected!
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:336:16)
gyp ERR! stack     at emitTwo (events.js:126:13)
gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
gyp ERR! System Darwin 17.2.0
gyp ERR! command "/usr/local/Cellar/node/8.9.1/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/machawk1/Downloads/station/node_modules/macos-alias
gyp ERR! node -v v8.9.1
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok 

> fs-xattr@0.1.17 install /Users/machawk1/Downloads/station/node_modules/fs-xattr
> node-gyp rebuild

gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
[...]

PR #519, opened 2017-12-04
Add further requirements to build on macOS

machawk1:
In #517 I ran into a barrier in building station. With @VictorBjelkholm's guidance, I was able to install the dependencies and get up-and-running. Adding this change to the README would help others get going quicker.


hacdias:
Thanks @machawk1!!!

PR #531, opened 2017-12-05
fix: update installer icons

hacdias:
That ^


hacdias:
Not needed for now. Take a look at this after deciding which builder use for electron.

PR #536, opened 2017-12-06
feat: hide on blur

### LINKED ISSUE #522: UI: Station popup should hide when context deselection is requested
In macOS, a common UX pattern is to select (full mousedown-then-mouseup) a menu item, which shows the popup even when the mouse is no longer pressed. Clicking elsewhere when this popup is open closes the popup. For example:

out

Station does not allow this "deselection" though. When the Station popup is open and somewhere outside of the popup is clicked, the popup remains visible. Selecting the Station icon again closes the menu but with most programs with a menu item, clicking outside of the popup is usually sufficient. This also means that while Station is displaying its popup, other menu item (e.g., Sound control) can be selected and displayed on top of the Station UI.

out

...which gets awkward. It would be more natural as a native application if the selection scheme followed that of the system. It might be useful to look at the windowing/display model used by other Electron apps with a menu item to see how they handle this pattern.

hacdias:
Closes #522.

image

I just don't know if the "package" icon is the best one. What is your opinion?

The icons we're using: [themify](http://themify.me/themify-icons).

/cc @machawk1 @diasdavid


machawk1:
@hacdias The package icon is very similar to the IPFS icon, which might cause confusion.

Even so, what is the icon supposed to convey? Is meant to mean "retain the view onblur", "hide the view onblur" or something else?

The pin icons from the themify would normally be very effective at conveying "retain this view" but per the other ticket, pinning is already in use in the IPFS nomenclature, so would lead to confusion.

hacdias:
It should mean that the window is going to be fixed so you can drag and
drop files to the interface directly.

machawk1:
@hacdias How about using ti-eye from that icon set instead of the "package" icon?

hacdias:
yeah, it can be:

image

my question is: to which point is this button self-explanatory?

machawk1:
I think there's still a UX issue in the placement. The + and folder icon add things to station whereas the heartbeat and eye-(aye) icon change the view. Maybe move the eye icon to be grouped with the heartbeat.

One way to make the button self-explanatory would be using highlighting to indicate an active state, e.g., increase the opacity or make the icon "whiter" instead of grey.  However, this view model is not used elsewhere in the UI as far as I can tell. Another way is to make the eye appear "closed" but that escapes it from the meaning of retaining the view onblur and frankly, all "closed eye" icons I've seen are difficult to visually decipher.

hacdias:
image

image

hacdias:
Could you check this one @diasdavid ?

hacdias:
@diasdavid can we merge this?

PR #546, opened 2017-12-17
Fix permissions + exit on error

VictorBjelkholm:
Changes the permission of the script, makes sure we exit directly when something fails and debug output for when things are OK.


hacdias:
Thanks!

PR #552, opened 2017-12-21
chore: add circleci

hacdias:


VictorBjelkholm:
Not sure we really need CircleCI for station. In general we're moving all CI to Jenkins, but still using CircleCI for the repositories where it already was activated. However, if we didn't use CircleCI, it might be better to not add it. We're just doing linting, which should be platform-independent anyways, and right now this just adds builds to our queue, making CircleCI overall slower for us.

hacdias:
@VictorBjelkholm sure thing. Can you disable it for me? It was @diasdavid who activated it and I don't think I have permission to change that.

hacdias:
What about AppVeyor @VictorBjelkholm ?

VictorBjelkholm:
@hacdias same thing there, it's just duplicating what we're already doing on jenkins + making the queues bigger. I'll disable both of them.

hacdias:
Thanks @VictorBjelkholm! I'll remove the files then. Can you take a look [here](https://github.com/ipfs-shipyard/station/pull/549#issuecomment-353320428)?

PR #562, opened 2017-12-22
Screenshot Shortcuts and Code Cleaning

hacdias:
- [x] Screenshot Shortcut
	- [x] CTRL/CMD-ALT-S
	- [x] Upload Screenshot
	- [x] Copy link to clipboard
	- Should there be a notification? (Notifications not working on Windows ). Or should we change the IPFS icon when we are uploading?
- [ ] ~Upload File Shortcut~
- [ ] ~Download selected hash shortcut~
- [x] Clean controls


hacdias:
@diasdavid AFAIK, Electron can't get selected text on screen that's not inside electron's view due to security reasons. The same happens to files: I can't know if the keys were pressed when selecting files so I don't think it is possible to create those shortcuts right now, but I'll investigate a bit more.

hacdias:
@diasdavid could you take a look at the code? I'm thinking of merging this and do the other two shortcuts for later (when it's possible or when I find a solution).

hacdias:
Absolutely hate Jenkins. Build failed on Windows, but it's working. It was a download error.

src/config.js
```diff
@@ -3,31 +3,39 @@ import path from 'path'
...
-  const p = path.resolve(path.join(__dirname, 'img'))
-
-  if (os.platform() === 'darwin') {
-    return path.join(p, 'icons/ice.png')
-  }
+// Set up what to do on Uncaught Exceptions
+process.on('uncaughtException', (error) => {
+  const msg = error.message || error
+  logger.error(`Uncaught Exception: ${msg}`, error)
+  dialog.showErrorBox('Uncaught Exception:', msg)
```
diasdavid: This error box should guide the user to the issue tracker (this repo) and recommend to open a issue with the error

src/config.js
```diff
@@ -3,31 +3,39 @@ import path from 'path'
...
+})
 
-  return path.join(p, 'ipfs-logo-ice.png')
-})()
+// Set up crash reporter or electron debug
+if (!isDev) {
+  require('electron').crashReporter.start({
+    productName: 'Station',
+    companyName: 'IPFS',
+    submitURL: 'https://ipfs.io',
```
diasdavid: What is this submitting to?

hacdias: This is submitting nothing to nowhere. I'll remove this. It seems to be here since the 'old days' 😛

src/config.js
hacdias: Will do this tomorrow or sunday.

PR #563, opened 2017-12-23
[WIP] refactor: use electron-menubar instead of menubar

hacdias:
I'm building a small module called 'electron-menubar' which is based on 'menubar', but with its code cleaned and a simpler API and a bunch of bugs fixed (like the correct position on Linux distributions).


diasdavid:
@hacdias that's rad! 👏🏽👏🏽👏🏽

PR #571, opened 2017-12-28
feat: pinning

hacdias:
Ref.: #518 

## Pin a new hash

[https://gyazo.com/cc95c17262a58052d43230cb756f5c09](https://gyazo.com/cc95c17262a58052d43230cb756f5c09)

# Search

[https://gyazo.com/7e1d694d77174bd6c3493812c11318ab](https://gyazo.com/7e1d694d77174bd6c3493812c11318ab)

# Animation while pinning

[https://gyazo.com/72e7b03963fa2eec270b2516f4f84437](https://gyazo.com/72e7b03963fa2eec270b2516f4f84437)

PR #572, opened 2017-12-28
refactor: menu

### LINKED ISSUE #560: Two buttons for the same thing
Right now we have two different buttons doing the same thing: turning the daemon on and off.

image

Should they be kept? IMO, we should remove the one from the bottom since it's repetitive. Although, it's more clear what it does than the one from the top right.

/cc @diasdavid

hacdias:
What do you think about something like this?

This:
- is more compact
- is more simple
- has easier navigation. 
- simplifies the process of adding new panes.

### Todo

- [x] Review setup screen
- [x] Light Theme
- [x] Review DOCS on Readme
- [x] Settings inside menubar

### Light Theme

image

image

image

image

### Dark Theme

image

image

image

image

This closes #560.

/cc @ipfs-shipyard/ipfs-gui-team


diasdavid:
Looking good @hacdias . Is this part of the work to make adding panes easier?

hacdias:
Yes, it will be easier to do so if this gets merged. Basically, on the left we have the available panes and it gets displayed on the right. 

Despites that, I think it's also easier for the end-user to navigate through the application instead of having those options in the bottom left as we have right now.

hacdias:
@diasdavid could you take a look at [this idea](https://github.com/ipfs-shipyard/ipfs-station/blob/debd4fc9716a21e853c45ac7c94fe6a2ec24b3c7/src/js/screens/menu.js), please?

- `panesOrder` is the order by which the options are showed on the menu.
- `panes` are the panes configuration.
- The `this` from the class is binded into the `render` function.

hacdias:
I'd also be in favour of having a lighter theme:

image

Or at least an option to toggle between dark and light. What are your thoughts?

hacdias:
Some still unanswered questions:

- Which should be the "default" pane? I mean, the one which opens by default when we start Station? The node information or files?
- Does it make more sense to have the "close station" button in Settings or in Info? In my opinion, it makes more sense in Settings.
- Related to that, we have #560. Should the 'Shutdown' button really close station? If so, we wouldn't need the previous option.

/cc @machawk1 @diasdavid 

The current state:

image

image

image

image

image

image

image

image

diasdavid:
> Which should be the "default" pane? I mean, the one which opens by default when we start Station? The node information or files?

The Informations Pane

> Does it make more sense to have the "close station" button in Settings or in Info? In my opinion, it makes more sense in Settings.

- Quit Station (as in close the app) - In the Settings
- Stop/Start the daemon - In the corner

> Related to that, we have #560. Should the 'Shutdown' button really close station? If so, we wouldn't need the previous option.

Shutdown/stop should just stop the daemon and not quit station.

Also, default to the dark theme but yes you can have the light theme.

image

image

Once this details are done, I suggest to go ahead and merge and focus on the Pinning feature #571 that will be a stellar feature.

diasdavid:
One more thing:

image

hacdias:
@diasdavid 

image

Should we keep both (circled in red)?

hacdias:
image

@diasdavid may I merge it?

diasdavid:
@hacdias go for it :)

PR #549, opened 2017-12-20
feat: js-ipfsd-ctl daemon-remote

hacdias:
> WIP

Depends on: https://github.com/ipfs/js-ipfsd-ctl/pull/176

- [x] Update on package.json ipfsd-ctl version.

Issue #542, opened 2017-12-08
Let Desktop use pre-existing IPFS installation or already running daemon


hacdias:
Ref.: https://github.com/ipfs/js-ipfsd-ctl/issues/182

/cc @machawk1 @diasdavid

machawk1:
It might also be good to advertise a way for other tools to leverage Station's IPFS installation. I am thinking of this in the context of https://github.com/oduwsdl/ipwb, which initially checks IPFS_PATH then defaults to ~/.ipfs if it hasn't been set. For other tools to leverage Station's IPFS, both the binary and the repo, might require an API on Station's end to ensure that Station gets all of the information needed for files added via other tools.

lidel:
Another data point from me :)

On one of my workstations I run go-ipfs under Docker.
Due to this my `~/.ipfs` has only one text file named  `~/.ipfs/api` with multiaddr of remote node:

```
$ cat ~/.ipfs/api
/ip4/127.0.0.1/tcp/5001
```

This setup seems to break startup of ipfs-desktop:

```$ DEBUG='desktop' npm start

> ipfs-desktop@0.3.0 start /home/lidel/projects/ipfs-desktop
> electron-forge start

✔ Checking your system
✔ Locating Application
✔ Preparing native dependencies: 3 / 3
✔ Launching Application
  desktop Application is ready +0ms
  desktop Starting daemon +5ms
  desktop Uncaught Exception: Error: ENOENT: no such file or directory, open '/home/lidel/.ipfs/config' +23ms
```

I guess ipfs-desktop should  detect `~/.ipfs/api` and connect to remote node 🚀

bedri:
Files added with command line ipfs (I think I had installed ipfs-go but it shouldn't matter which client user uses) does not show up in the ipfs-desktop GUI. I guess this should also go under this topic. If requested I can open a separate issue for this.

Thank you.

hacdias:
@bedri the files we show on IPFS Desktop are the same we show on IPFS WebUI: from the Mutable File System. They should be added through `ipfs files write` and not `ipfs add`.

bedri:
@hacdias oh ok. I did not know that. I will check it.
Although I am able to see the files that I add with "ipfs add" from Web UI. I am confused at that part.
I couldn't also managed to use ipfs files write to add files but i will figure it out eventually.

hacdias:
[...]

PR #585, opened 2018-01-10
fixes: use debug, correctly open settings and offline stats

hacdias:
- [x] Use `debugger` instead of Winston.
- [x] Open the settings with `openItem` instead of `openExternal`.
- [x] Fixes offline stats.

**Edit:** just noticed that the daemon is always exiting with error on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "main": "src/index.js",
   "dependencies": {
     "auto-launch": "^5.0.5",
+    "debug": "^3.1.0",
     "electron-compile": "^6.4.2",
     "electron-is-dev": "^0.3.0",
     "electron-menubar": "^1.0.1",
@@ -23,8 +24,7 @@
     "react": "^16.2.0",
     "react-dnd": "^2.5.4",
     "react-dnd-html5-backend": "^2.5.4",
-    "react-dom": "^16.2.0",
-    "winston": "^3.0.0-rc1"
+    "react-dom": "^16.2.0"
   },
   "devDependencies": {
     "babel-eslint": "^8.2.1",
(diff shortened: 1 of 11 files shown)


hacdias:
@VictorBjelkholm can you try running this branch to see if you "Node Settings" already work? 😄

VictorBjelkholm:
I downloaded the .deb from https://ci.ipfs.team/blue/organizations/jenkins/IPFS%20Shipyard%2Fipfs-desktop/detail/PR-585/4/artifacts

First problem I hit was that I was running go-ipfs since before and ipfs-station just refused to start up. It should reuse the already running daemon if it exists.

Second issue I hit was that all the values were undefined since I wasn't connected to the internet. It should still list the Node ID, even though I cannot reach the internet.

Thirdly, "Node Settings" does open up a editor for editing the settings now.

hacdias:
@VictorBjelkholm 

> First problem I hit was that I was running go-ipfs since before and ipfs-station just refused to start up. It should reuse the already running daemon if it exists.

Ref.: #542. Will be easier after #549 gets merged.

> Second issue I hit was that all the values were undefined since I wasn't connected to the internet. It should still list the Node ID, even though I cannot reach the internet.

Gonna check this tomorrow. 😄 

> Thirdly, "Node Settings" does open up a editor for editing the settings now.

Awesome!

> Not sure ^ is new things, but things I noticed was broken in the build for this PR.

Sometimes the build just fails... I'll restart it.

hacdias:
@VictorBjelkholm btw, you can now run this with `DEBUG=desktop yarn start` to see the log. Could you check if you can find out the reason why IPFS is being killed now?

PR #587, opened 2018-01-11
feat: open on click

### LINKED ISSUE #583: Open files on click

hacdias:
Closes #583.

This removes the option to copy a file's hash to the clipboard and instead it opens the link in your default browser.

### DIFF
--- a/src/controls/main/index.js
+++ b/src/controls/main/index.js
@@ -3,6 +3,7 @@ import downloadHash from './download-hash'
 import fileHistory from './file-history'
 import menuShortcuts from './menu-shortcuts'
 import openFileDialog from './open-file-dialog'
+import openUrl from './open-url'
 import openWebUI from './open-webui'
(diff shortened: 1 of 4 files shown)


diasdavid:
> This removes the option to copy a file's hash to the clipboard and instead it opens the link in your default browser.

The goal was not to remove but to rather have both buttons. //cc @pgte

hacdias:
@pgte @diasdavid now it does both: if you click on 'Copy link' it.. well.. copies the link. If you click outside of it, will open the link in your browser 😄

diasdavid:
@hacdias why not two buttons?

hacdias:
@diasdavid I think it is more natural to click on something to open it. Don't you think so?

hacdias:
@diasdavid two buttons: open and copy link 😄

hacdias:
image

diasdavid:
Thank you @hacdias :)

PR #591, opened 2018-01-11
Update README.md

flyingzumwalt:
switches wording install section to put the instructions first rather than a reference to something that's not working yet.


hacdias:
Thank you @flyingzumwalt

PR #592, opened 2018-01-12
Enhance your calm, README.

olizilla:
Makes the readme a little calmer. 

- Add a static screenshot to retain the visual interest without the animation.
- Pull the gifs lower down, and make them a **Usage** section. 
- Drop the settings gif and the light theme gif.
- Re-order the content as per standard-readme.
- Focus the install section on users. Clarify that installing from source is there if you need it. Simplify the content.
- Group the dev info under a **Developer Guide** so users get a hint that it's ok to ignore it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 <h1 align="center">
-  <a href="https://ipfs.io">IPFS Desktop</a>
+  IPFS Desktop
 </h1>
 
-<h3 align="center">A menubar IPFS application to get you on the Distributed Web!</h3>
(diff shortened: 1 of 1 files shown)


olizilla:
I find it much easier to think about this change by viewing the finished readme rather than trying to review the github dif, so jump here https://github.com/ipfs-shipyard/ipfs-desktop/blob/5ff9dd6726abb762329f31fc05d9bafe4eeb735c/README.md if you feel the same

hacdias:
Can you take a look at this @diasdavid?

PR #596, opened 2018-01-16
Fixes Bunch

hacdias:
## Related:

- ~Initial Issue: https://github.com/ipfs-shipyard/ipfs-webui/issues/616~
- ~IPFS WebUI PR: https://github.com/ipfs-shipyard/ipfs-webui/pull/618~
- ~[`ipfs-react-components`](https://github.com/hacdias/ipfs-react-components) - common React Components that are used on WebUI and Desktop.~
- Read more on https://github.com/ipfs-shipyard/pm-ipfs-gui/issues/29

## Goals

- [x] Make components as independent as possible
- [x] Remove repo.lock when not running (fix #588)
- [x] Fix #590 by using MFS
- [x] Save pins on MFS (ref.: #597)
- [x] Order addresses on info pane (fix #599)
- [x] Show previously pinned stuff (fix #600)
- [x] Save Screenshots on MFS
- [x] While daemon is turned off, let the user turn it on or close Desktop only.
- [x] Improve Files pane
- [x] Add some basic tests to start the testing journey #50

Issue #50, opened 2015-10-05
Add test setup


dignifiedquire:
> Every decent (software) project needs tests
## "Frontend" (browser processes)

This is about testing react components in isolation. To run the tests we should use [karma](http://karma-runner.github.io/0.13/index.html) and spawn a chrome process (close enough to electrons chromium instances). Interesting modules and notes
- https://github.com/sheepsteak/react-shallow-testutils
- https://github.com/percyhanna/rquery
- https://github.com/jhnns/rewire-webpack (mock require calls, especially ipc)
- https://github.com/pgte/nock (mock http calls)
## "Backend" (electron main process)

This is much harder, as too much functionality depends on running the real deal aka an electron instance. So very happy to hear any suggestions around this.

hacdias:
Spectron --> https://github.com/electron/spectron

hacdias:
**State of the thing:** Taking [this branch](https://github.com/ipfs-shipyard/ipfs-desktop/tree/feat/components/src) as the state from which we should start writing tests, we've got many different components to test:

- **components** which Simple React components which seem to be relatively easy to test
- **controls** which are subdivided in two:
	- **main** process handlers.
	- **renderer** process handlers.
- **panes** which are a bit more complex React Components that can use Electron's API directly.
- **utils** some simple... utils? Easy to test.
- **views** main entry points for the pages Welcome and Menubar.
- ```index.js```, `errors.js`, `config.js`

I can easily start to create the tests for the components and the utils. Although, I'm not sure how to do testing for the controls, which are basically event handlers of `ipcMain` and some other general stuff. Nor for panes and views.

olizilla:
WIP on adding spectron test in https://github.com/ipfs-shipyard/ipfs-desktop/pull/730

hacdias:
Superseded by #1121

PR #604, opened 2018-01-27
Start addressing the tests

hacdias:
Ref.: #50 

Starting by testing the components directory.

This uses `feat/components` as base. Can only be merged after #596


hacdias:
With the current configuration, I'm getting this:

```
----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
All files |  Unknown |  Unknown |  Unknown |  Unknown |                |
----------|----------|----------|----------|----------|----------------|
```

@VictorBjelkholm do you think you could help with this? `nyc` doesn't seem to be working correctly. I added the configuration to `package.json` and the `instabul` plugin to Babel.

VictorBjelkholm:
I think you should start with setting up tests with spectron/acceptance testing. Main idea is to just make sure the core UI functionality works. Then you can start adding unit tests for making sure every detail is working. About coverage, we still working on making sure the secret management in Jenkins is stable and secure. So not gonna do that from jenkins just yet.

Another point, in general you should avoid having files with big and small letters, as you'll hit confusing errors when you name some file `HelloWorld.md` and on the OS you're developing you can load it as `helloworld.js` but it won't work on others (or vice-versa). It's  easier to always have lower-case naming and use underscore instead.

hacdias:
> I think you should start with setting up tests with spectron/acceptance testing. Main idea is to just make sure the core UI functionality works.

I'd like to do that but it isn't simply working... yet! Our electron is using some pre-compilers for LESS and Babel and I wasn't still able to decipher how to use that with Spectron.

> Another point, in general you should avoid having files with big and small letters...

I'm just using CamelCase for files with components to follow React's best practices, but I can change that. Should I?

hacdias:
@VictorBjelkholm I'm trying to make this simply start the app and nothing seems to happen. I'll increase the timeout for something like 10 mins and see what happens.

image

Ref.: https://github.com/electron-userland/electron-compile/issues/178

hacdias:
Many conflicts. 

- [ ] Fix after #596 gets merged.

hacdias:
This tests got merged on #596. Keep working on #596

PR #606, opened 2018-01-28
fix: refresh pins and stats at better rates

### LINKED ISSUE #600: Pinned hashes don't show up
I have just started using IPFS Desktop, and noticed that my pinned hashes weren't listed in the Pin tab. But when I run `ipfs pin ls` in the command line I get several hashes, including non-default ones. These hashes were pinned before I used IPFS desktop, but shouldn't it still see them?
EDIT: I'd like to add that hashes pinned through IPFS desktop show up in the command line, just not the other way around. I'd like to also say that when I add a file, it doesn't show in the pin tab as well, even though files you add are automatically pinned.

hacdias:
Fixes #600.

Ping @Cole128

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "electron-squirrel-startup": "^1.0.0",
     "file-extension": "^4.0.1",
     "go-ipfs-dep": "^0.4.13",
-    "ipfs-stats": "^1.2.3",
+    "ipfs-stats": "^1.2.4",
     "ipfsd-ctl": "^0.27.0",
     "is-ipfs": "^0.3.2",
     "moment": "^2.20.1",
--- a/src/controls/main/stats.js
+++ b/src/controls/main/stats.js
@@ -60,7 +60,11 @@ export default function (opts) {
   events.on('node:started', () => {
     debug('Configuring Stats Poller')
 
-    poller = new StatsPoller(ipfs(), 1000)
+    poller = new StatsPoller(ipfs(), {
+      all: 3 * 1000,
+      peers: 5 * 1000,
+      repo: 30 * 60 * 1000
+    })
 
     if (menubar.window && menubar.window.isVisible() && polling.length !== 0) {
       poller.start(polling)
--- a/src/screens/menubar.js
+++ b/src/screens/menubar.js
@@ -86,6 +86,16 @@ class Menubar extends Component {
     return this.listeners[key]
   }
 
+  onPinned = (event, pinset) => {
+    this.setState({pinned: pinset})
+
+    setTimeout(() => {
+      if (this.state.route === 'pinned') {
+        ipcRenderer.send('request-pinned')
+      }
+    }, 5000)
+  }
+
   _changeRoute = (route) => {
     if (route === this.state.route) return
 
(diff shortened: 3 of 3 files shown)


Cole128:
@hacdias Looks good 😄 , although will indirect pins show up? And how do I install it for my mac since it hasn't been packaged as a dmg yet?

hacdias:
@Cole128 `indirect` pins won't show up. The only thing I added was the refreshing at every one sec. 

About the .dmg:

image

When one of those little guys gets a check mark saying the build passed, you can click on Details --> Artifacts --> And there should be your dmg 😄

Cole128:
@hacdias Great! Once I install it and test it out I'll let you know whether it works.

Cole128:
@hacdias I can't open the dmg, it says image not recognized.

hacdias:
Could you try on the other link? I don't have a macOS and I can't really help with that but please try the other link. Otherwise, tomorrow I can try to make it rebuild the image.

Cole128:
@hacdias I tried both.

VictorBjelkholm:
@Cole128 could you please comment with the exact link you downloaded the .dmg from and also what macOS version you're running?

Cole128:
@VictorBjelkholm https://ci.ipfs.team/job/IPFS%20Shipyard/job/ipfs-desktop/job/update-pin/1/artifact/out/make/IPFS%20Desktop-0.3.0.dmg
and https://ci.ipfs.team/job/IPFS%20Shipyard/job/ipfs-desktop/job/PR-606/1/artifact/out/make/IPFS%20Desktop-0.3.0.dmg

I am running macOS High Sierra, version 10.13.2 (17C88)

hacdias:
@Cole128  Could you try [this one](https://ci.ipfs.team/job/IPFS%20Shipyard/job/ipfs-desktop/job/PR-606/2/artifact/out/make/IPFS%20Desktop-0.3.0.dmg)?

---

Just as a sidenote: builds are failing due to https://github.com/sinonjs/nise/issues/27

lgierth:
Calling `pin/ls` once per second can cause quite the load on go-ipfs and the harddrive. Is it really neccessary to have this happen every second?

If live updates are neccessary, we could instead emit `pin-added` and `pin-removed` events (we might already be doing that), which ipfs-desktop could listen to.

lgierth:
> we could instead emit pin-added and pin-removed events

To clarify, I mean the `ipfs log` command.

hacdias:
@lgierth didn't know about that!! Thanks. I'll check that out and see what if there are any events related to the pins.

hacdias:
@lgierth yeah, I've noticed that before lunch and I'm planning to change it to only call the command again one second after having the other's response.

How should I listen to those events? Using `ipfs.log.tail([callback])`? It seems a bit impractical since it doesn't allow me to handle specific event types.

lgierth:
Mh it looks like the `ipfs log` command is quite odd -- it seems like `ipfs log tail` lets deals with events, but `ipfs log ls` and `ipfs log level` deal with log output. (Events and log output are two different things).

Maybe a filter argument to `ipfs log tail` is neccessary, or just moving the events stuff into its own command.

hacdias:
@Cole128 [here's a new dmg](https://ci.ipfs.team/job/IPFS%20Shipyard/job/ipfs-desktop/job/PR-606/5/artifact/out/make/IPFS%20Desktop-0.3.0.dmg)!

Cole128:
@hacdias I still get the error: Image not recognized.

hacdias:
@Cole128 unfortunately, I don't. Maybe @lgierth?

hacdias:
I'll merge this for now and then we can try to check out what's happening with the DMG.

PR #624, opened 2018-03-21
fix: upload directory with subdirectories [WIP]

### LINKED ISSUE #613: Cannot add folder more than a single level deep
Adding a folder with a sub folder with a file 

```
parent
└── sub
    └── two.txt
```

leaves me with the upload progress bar. I can still click around the app. It feels like the upload is in progess, just taking a really really long time, like > 10 minutes to upload a 200 byte file that happens to be 2 dirs deep.

hacdias:
Fixes #613.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -9,6 +9,8 @@ import KeyValueStore from './utils/key-value-store'
 
 const debug = dbgger('desktop')
 
+process.setFdLimit(2048)
+
 function logo (color) {
   const p = path.resolve(path.join(__dirname, 'img'))
 
--- a/src/controls/utils/upload-files.js
+++ b/src/controls/utils/upload-files.js
@@ -1,56 +1,41 @@
 import path from 'path'
-import fs from 'fs'
 
 function join (...parts) {
   const replace = new RegExp('/{1,}', 'g')
   return parts.join('/').replace(replace, '/')
 }
 
-function clean (files, root) {
-  const res = []
+let adding = 0
 
-  files.forEach((file) => {
-    const stat = fs.lstatSync(file)
-    const dst = join(root, path.basename(file))
+async function add (files, root, ipfs) {
+  for (const file of files) {
+    const res = await ipfs().add([file], {recursive: true, wrap: true})
+    const f = res[res.length - 1]
+    const src = `/ipfs/${f.hash}`
+    const dst = join(root, path.basename(f.path))
 
-    if (stat.isDirectory()) {
-      const files = clean(fs.readdirSync(file).map(f => path.join(file, f)), dst)
-      res.push({dir: true, dst: dst}, ...files)
-    } else {
-      res.push({
-        dst: dst,
-        src: file
-      })
-    }
-  })
-
-  return res
+    await ipfs().files.cp([src, dst])
+  }
 }
 
 export default function uploadFiles (opts) {
   let {ipfs, debug, send} = opts
-  let adding = 0
 
   const sendAdding = () => { send('adding', adding > 0) }
   const inc = () => { adding++; sendAdding() }
   const dec = () => { adding--; sendAdding() }
 
+  const anyway = () => {
+    dec()
+    send('files-updated')
+  }
+
   return (event, files, root = '/') => {
     debug('Uploading files', {files})
-    files = clean(files, root)
-
     inc()
-    Promise.all(files.map(file => {
-      if (file.dir) {
-        return ipfs().files.mkdir(file.dst)
-      }
 
-      return ipfs().files.write(file.dst, file.src, {create: true})
-    })).then(() => {
-      dec()
-      send('files-updated')
-    }).catch((e) => {
-      debug(e.stack)
-    })
+    add(files, root, ipfs)
+      .then(anyway)
+      .catch((e) => { anyway(); debug(e.stack) })
   }
 }
--- a/src/panes/Files.js
+++ b/src/panes/Files.js
@@ -24,7 +24,7 @@ const fileTarget = {
       filesArray.push(files[i].path)
     }
 
-    ipcRenderer.send('drop-files', filesArray, component.state.root)
+    ipcRenderer.send('drop-files', filesArray, component.props.root)
   }
 }
 


lidel:
Just tested it, adding 2-3 layers of files works.. fine. 

But I think I've found kinda unrelated bug when adding bigger directory trees: 

Tested against revision 3ded93ecbd490e498bda3031fe3c200cb4ead47c
To replicate, just try to add `ipfs-desktop` sources with `.git` via "add directory" icon.
Do you get the same error as below?

screenshot_2


May be that my system has low limits, but worth addressing at some point.
Debug log from adding `ipfs-desktop` on my box:
```
$ DEBUG='desktop' yarn start
$ electron-forge start
✔ Checking your system
✔ Locating Application
✔ Preparing native dependencies: 3 / 3
✔ Launching Application
  desktop Application is ready +0ms
  desktop Starting daemon +4ms
  desktop Daemon started +3s
  desktop Configuring Stats Poller +0ms
  desktop TypeError: stream.pipe is not a function
  desktop     at Promise (/home/lidel/project/ipfs-desktop/src/controls/main/pinned-files.js:30:12)
  desktop     at Promise (<anonymous>)
  desktop     at collect (/home/lidel/project/ipfs-desktop/src/controls/main/pinned-files.js:29:10)
  desktop     at ipfs.pin.ls.then.then.then.res (/home/lidel/project/ipfs-desktop/src/controls/main/pinned-files.js:76:56)
  desktop     at <anonymous>
  desktop     at process._tickCallback (internal/process/next_tick.js:188:7) +105ms
  desktop Uploading files { files: [ '/home/lidel/project/ipfs-desktop' ] } +12s
  desktop Uncaught Exception: Error: connect EMFILE 127.0.0.1:37895 - Local (undefined:undefined)
  desktop     at Object.exports._errnoException (util.js:1024:11)
  desktop     at exports._exceptionWithHostPort (util.js:1047:20)
  desktop     at internalConnect (net.js:934:16)
  desktop     at net.js:1033:9
  desktop     at _combinedTickCallback (internal/process/next_tick.js:131:7)
  desktop     at process._tickCallback (internal/process/next_tick.js:180:9) +16s
```

hacdias:
@lidel as far as I've tried, there is some kind of limit and I didn't understand it yet because I remember asking someone and there wasn't supposed to be one. I'll try to replicate.

hacdias:
@lidel trying on linux what happened was: 

1. Tried to upload
2. Nothing apparently happened
3. Tried to upload again
4. Error `file already exists`
5. Refreshed Desktop and found out that it was addedd on the first time.

What I think is that there is some problem updating the pane. I'll solve it 😄

Although the `EMFILE` error I have never seen it before. I have no idea of what it can be. For what I read, EMFILE means `too many open files`. Maybe I should not try to add everything at once but like 10 each time to avoid that error.

hacdias:
@lidel @olizilla I think https://github.com/ipfs-shipyard/ipfs-desktop/pull/624/commits/b43edb07c690a11bd7fff84e1874deebe2108ccb actually solves this problem. I created a `async` function which uploads everything synchronously, file by file, directory by directory.

The `file does not exist` error was created because since I was using Promise.all, it could eventually try to upload files to directories that didn't exist yet.

I ask you to try this out please and let me know if it works 😄

lidel:
I got you some quick feedback before going to  :sleeping_bed: :zzz: 
Performed my usual test (adding `ipfs-desktop` project tree ~600MB) and good news is that the error changed, and problems moved to bigger directory trees. 

I got this `Error: blockservice is closed` perk after about 3 minutes:   [log1.txt](https://github.com/ipfs-shipyard/ipfs-desktop/files/1839665/log1.txt).
After I start it  again I see the ipfs-desktop is only 64MB in size, so it seems app stopped adding stuff at that offset. I was not able to replicate it again tho.

I tried again, and this time got `ENOENT: no such file or directory` around 240MB:  [log2.txt](https://github.com/ipfs-shipyard/ipfs-desktop/files/1839668/log2.txt). This may be a false-positive as I actually do not have this file on my disk (possibly that electron changed dir tree since adding started?).

For the final test I've added my `ipfs-companion` project dir instead (~845MB as it includes different versions of Firefox), and after >1h it ended with >600MB added and `ECONNREFUSED` error: 
[log3.txt](https://github.com/ipfs-shipyard/ipfs-desktop/files/1839805/log3.txt).

Hope this helps :)

hacdias:
I think this was it. @olizilla gave me the idea of using `ipfs.add` and then `ipfs.files.cp` to the MFS because `ipfs.add` supports adding directories directly. 

Although, if I add a dir using `ipfs.add` it returns an array with the hash of every added file. For what I saw, the root directory is always the last element, but I'd like to know if you can confirm that.

olizilla:
This works for my small test case that is 2 dirs deep with 3 files.

Trying @lidel's test case of adding the ipfs-companion src tree (with node_modules) causes

```
  desktop Uploading files { files: [ '/Users/oli/Desktop/foo' ] } +28s
  desktop Uploading files { files: [ '/Users/oli/Code/ipfs-shipyard/ipfs-companion' ] } +1m
  desktop Error: connect EMFILE 127.0.0.1:57379 - Local (undefined:undefined)
  desktop     at Object.exports._errnoException (util.js:1024:11)
  desktop     at exports._exceptionWithHostPort (util.js:1047:20)
  desktop     at internalConnect (net.js:934:16)
  desktop     at net.js:1033:9
  desktop     at _combinedTickCallback (internal/process/next_tick.js:131:7)
  desktop     at process._tickCallback (internal/process/next_tick.js:180:9) +7s
  desktop Uncaught Exception: Error: EMFILE: too many open files, open '/Users/oli/Code/ipfs-shipyard/ipfs-companion/node_modules/choo/example/node_modules/browserify/test/multi_entry_cross_require/c.js' +1s
```

where 
```
  desktop Uploading files { files: [ '/Users/oli/Desktop/foo' ] } +28s
```
is the simple test case working, which is new and good news ✨ but then there is a new error with the `/Users/oli/Code/ipfs-shipyard/ipfs-companion` which is a has many files.

It then continues to try and add all the files, but each one throws a new `too many open files` error, so now I've got hundreds of error alert boxes to close 💥

lidel:
Now I get the same error as @olizilla (`EMFILE: too many open files` for `ipfs-companion` tree). 

But it is a good news, means we are just missing some way to limit the number of file handlers used in parallel during the add. 

Seems that Beaker/Dat team had similar problem and `graceful-fs` helped: https://github.com/datproject/dat/pull/570

hacdias:
@lidel yeah, I'll merge this and open an issue on js-ipfs-api after getting a reproducible example.

lidel:
Hm.. this does not seem to make any difference, same as https://github.com/ipfs-shipyard/ipfs-desktop/pull/624#issuecomment-375619973

hacdias:
Hmm. Can you try adding it manually via a script using jsipfsapi? I can't reproduce the EMFILE error in any Linux virtual machine I have.

hacdias:
I merged this and I'll open a new issue for the EMFILE errors.

PR #655, opened 2018-08-28
Remove Pinning and add Add By Path

### LINKED ISSUE #652: Remove the pinning feature
Some time ago (one year ago more or less), the WebUI moved from Pinning to using the MFS entirely.

Since we're going to incorporate the new WebUI in the following months on Desktop, I also suggest removing the Pinning feature from it so the users start getting more comfortable with MFS.

Feedback needed: @diasdavid @ipfs-shipyard/gui

hacdias:
Closes #652


src/controls/main/files.js
```diff
@@ -83,13 +83,31 @@ function move (opts) {
   }
 }
 
+function addByPath (opts) {
+  const { ipfs } = opts
+
+  return async (_, hash, path) => {
+    if (!hash.startsWith('/ipfs')) {
```
lidel: People may try to add `/ipns/` paths, and that should not be prefixed.

hacdias:
@lidel I changed the icon to the pinning one just to make it more familiar for now. What do you think?

src/panes/Files.js
```diff
@@ -95,6 +96,38 @@ class Files extends Component {
...
+
+    if (!validateIPFS(hash)) {
+      // es-no-alert
+      window.alert('Invalid IPFS Path!')
+      return
+    }
+
+    const path = await prompt({
+      title: 'Add by IPFS Path',
+      label: 'How do you want to tag it?',
```
lidel: Now that we are in Files  "How do you want to name it?"  is a better label.

src/panes/Files.js
```diff
@@ -95,6 +96,38 @@ class Files extends Component {
     return <Breadcrumbs navigate={navigate} path={this.props.root} />
   }
 
+  addFromIPFS = async () => {
+    const prompt = require('electron-prompt')
+
+    const hash = await prompt({
+      title: 'Add by IPFS Path',
+      label: 'Write the IPFS path to add:',
```
lidel: "Enter the IPFS path to add:"

src/panes/Files.js
```diff
@@ -144,6 +177,7 @@ class Files extends Component {
             <IconButton active={this.state.sticky} onClick={this.toggleStickWindow} icon='eye' />
 
             <div className='right'>
+              <IconButton onClick={this.addFromIPFS} icon='pin-alt' />
```
lidel: `IconButton` should accept `title` as one of its props, and use it like this:
```html
<button title="Add by IPFS Path">[icon]</button>
```

We really need a tooltip that describes what the button does :)

hacdias:
Done @lidel 😄

src/controls/main/pinned-files.js
```diff
@@ -1,146 +1,51 @@
...
-      .then((res) => {
-        for (const hash in res) {
-          if (hash in pins) {
-            pins[hash] = res[hash]
-          }
-        }
+  try {
+    await ipfs().files.stat(PATH)
+  } catch (e) {
+    if (e.toString().indexOf('file does not exist') !== -1) {
```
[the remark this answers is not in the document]
hacdias: Thanks! Merging now 😄

lidel:
@hacdias would be good to make a new release with this fix.
New users would not need to see migrated pinset when new WebUI lands :)

hacdias:
@lidel I'd like to but Jenkins isn't building for Windows and there seems to be an issue with macOS I can't replicate (#656)... Doesn't seem to be a good time for release 😞

PR #694, opened 2018-11-17
fix: do not add CORS Credentials header

lidel:
AFAIK IPFS Desktop does not need this header.

See also:

- Potential problem with  `Access-Control-Allow-Credentials`: https://github.com/ipfs/go-ipfs/issues/5745
- HTTP Headers Cleanup: https://github.com/ipfs/in-web-browsers/issues/132


hacdias:
Thank you @lidel 😄 I'm merging this!

PR #696, opened 2018-11-19
feat: automatically migrate repository (fixes #687)

### LINKED ISSUE #687: Hangs if  repo requires an update (missing flag for automatic migration)
> Issue found while testing "The PR" @  https://github.com/ipfs-shipyard/ipfs-desktop/pull/662#issuecomment-437576673

## Bug

If `IPFS_PATH` points at repo created with go-ipfs in version older than one bundled with ipfs-desktop, it will hang and never show tray icon:

<details>
  <summary>Click to expand debug log</summary>

```console
$ DEBUG='*' npm run start
(...)
winston:file logged 717 2018-11-12T11:41:25.385Z info: Starting IPFS { type: 'go', flags: [ [length]: 0 ], path: '/home/lidel/.ipfs' }

  ipfsd-ctl:daemon:stdout Initializing daemon...
  ipfsd-ctl:daemon:stdout go-ipfs version: 0.4.18-
  ipfsd-ctl:daemon:stdout Repo version: 7
  ipfsd-ctl:daemon:stdout System version: amd64/linux
  ipfsd-ctl:daemon:stdout Golang version: go1.11.1
  ipfsd-ctl:daemon:stdout Found outdated fs-repo, migrations need to be run.
  ipfsd-ctl:daemon:stdout Run migrations now? [y/N] +0ms
```

</details>

## Fix

The fix is to enable automatic migrations by always starting  `ipfs daemon` with `--migrate=true` parameter.

[This is what Docker image is already doing](https://github.com/ipfs/go-ipfs/blob/93c4f19169e0570adfb35fb6e9420d414c64a733/Dockerfile#L81).

hacdias:
Adds flag `--migrate=true` to migrate the repo automatically to the newest version.

License: MIT

### DIFF
--- a/src/utils/daemon.js
+++ b/src/utils/daemon.js
@@ -9,6 +9,10 @@ export default async function createDaemon (opts) {
   opts.flags = opts.flags || []
   opts.keysize = opts.keysize || 4096
 
+  if (!opts.flags.includes('--migrate=true')) {
+    opts.flags = [...opts.flags, '--migrate=true']
+  }
+
   if (opts.type !== 'go') {
     throw new Error(`${opts.type} connection is not supported yet`)
   }
--- a/src/utils/daemon.js
+++ b/src/utils/daemon.js
@@ -6,12 +6,8 @@ import logger from './logger'
 export default async function createDaemon (opts) {
   opts.type = opts.type || 'go'
   opts.path = opts.path || ''
-  opts.flags = opts.flags || []
+  opts.flags = opts.flags || ['--migrate=true', '--routing=dhtclient']
   opts.keysize = opts.keysize || 4096
-
-  if (!opts.flags.includes('--migrate=true')) {
-    opts.flags = [...opts.flags, '--migrate=true']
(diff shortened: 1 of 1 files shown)


src/utils/daemon.js
```diff
@@ -9,6 +9,10 @@ export default async function createDaemon (opts) {
   opts.flags = opts.flags || []
   opts.keysize = opts.keysize || 4096
 
+  if (!opts.flags.includes('--migrate=true')) {
```
olizilla: This seems like a suprising pattern. If you pass in an flags, this code will modify it. It'd be less suprising if we add `--migrate=true` in the default flags array. If the caller chooses to modify it then they should pass all the flags they want.

```js
opts.flags = opts.flags || ['--migrate=true']
```

hacdias: Yes, you're right. Will add that and `--routing=dhtclient` that @lidel mentioned. That one got lost during revamp haha

PR #702, opened 2018-11-19
wip: remove jenkins, add travis

hacdias:


olizilla:
This looks good, what else needs to happen here?

olizilla:
I don't know  if we need all this but here is [electon-builders sample travis.yaml](https://www.electron.build/multi-platform-build#sample-travisyml-to-build-electron-app-for-macos-linux-and-windows)

```yaml
matrix:
  include:
    - os: osx
      osx_image: xcode9.4
      language: node_js
      node_js: "10"
      env:
        - ELECTRON_CACHE=$HOME/.cache/electron
        - ELECTRON_BUILDER_CACHE=$HOME/.cache/electron-builder

    - os: linux
      services: docker
      language: generic

cache:
  directories:
    - node_modules
    - $HOME/.cache/electron
    - $HOME/.cache/electron-builder

script:
  - |
    if [ "$TRAVIS_OS_NAME" == "linux" ]; then
      docker run --rm \
        --env-file <(env | grep -iE 'DEBUG|NODE_|ELECTRON_|YARN_|NPM_|CI|CIRCLE|TRAVIS|APPVEYOR_|CSC_|_TOKEN|_KEY|AWS_|STRIP|BUILD_') \
        -v ${PWD}:/project \
        -v ~/.cache/electron:/root/.cache/electron \
        -v ~/.cache/electron-builder:/root/.cache/electron-builder \
        electronuserland/builder:wine \
        /bin/bash -c "yarn --link-duplicates --pure-lockfile && yarn release --linux --win"
    else
      yarn release
    fi
before_cache:
  - rm -rf $HOME/.cache/electron-builder/wine

branches:
  except:
    - "/^v\\d+\\.\\d+\\.\\d+$/"
```

notabley this is using docker and wine on linux to build the windows binaries, and our much simpler travis file seems to be building just fine without all that complexity, but it's worthing keeping the electron-builder suggestion around in case we hit any problems

hacdias:
@olizilla it seems to work, but it's just because we don't have a secret stored. Travis for Windows still has that issue so we will actually need that **big** travis file to make it work.

hacdias:
> Re: https://github.com/ipfs-shipyard/ipfs-desktop/pull/744#issuecomment-446904931

hacdias:
@olizilla while everything seems ok and working, the truth is that it isn't! Building Windows on Docker packs up with the IPFS binary for linux and not for Windows so it doesn't work. Shall I add AppVeyor while Travis for Windows isn't reliable?

hacdias:
Actually that might happen due to caching the dependencies. Lemme try removing that.

hacdias:
@olizilla nope, we really need to build it on Windows...

olizilla:
@hacdias is this the only docs for the travis window secrets issue https://travis-ci.community/t/windows-instances-hanging-before-install/250/12 ?

Options are, manually add windows binaries to the release, you can still build it on CI, you just need to be able to copy the installer over to GH, or set up  an additional CI for windows.

hacdias:
@olizilla this is [the official acknowledgment](https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264).

Like, it is building on Linux, but with the wrong IPFS binaries. We would need to use Jenkins to produce the correct Windows binaries. But then we would need to upload them manually.

I'll try out App Veyor.

hacdias:
@olizilla the current configuration will do the following:

If **there is a draft** and **is building on a CI**, it will add the binaries to the draft. If there is no draft, nothing will be done. We can use this to prevent being always uploading binaries on our builds.

We can change this later with channels.

hacdias:
Merging this. Then we can add  more features. I drafted a release and the binaries will show up there soon enough. We can release it tomorrow 😄

PR #706, opened 2018-11-23
feat: i18n

### LINKED ISSUE #650: Make it i18n friendly
@satazor implemented this for js.ipfs.io using react-intl and it is super easy to work with. @satazor, wanna shine some light here?

hacdias:
Basically done. The locales are on `src/locales`. @lidel would you help me by setting up Transifex for IPFS Desktop? 😄 

closes #650


hacdias:
@lidel maybe the language detector we use for Web UI can't detect the correct language of the system. I'll try out on my computer but I believe that a change on preload script will be able to solve that :)

hacdias:
@lidel merged. I will check out the Web UI issue now!

PR #708, opened 2018-11-24
fix: Web UI language matches menubar

hacdias:
This should fix the issue where the Web UI language wouldn't match menubar's. @lidel could you take a look to see if it works on your end?

License: MIT


lidel:
It detected `LANGUAGE=pl` in my shell environment and works as expected!
(I need to improve  Polish translation of Web UI tho :))

PR #717, opened 2018-11-27
Remove malicious package from transitive dependencies

lidel:
TL;DR  A package (flatmap-stream) got updated to include some dodgy code relating to crypto currency hijacking. This got included in event-stream which was pulled in by npm-run-all.

Context: https://github.com/dominictarr/event-stream/issues/116

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1007,7 +1007,7 @@
         },
         "hoek": {
           "version": "4.2.1",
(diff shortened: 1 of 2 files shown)


olizilla:
Thanks for the quick action @lidel. I've bumped the `package-lock.json`. The one generated from `npm@6.4.1` updates a bunch of `resolved` urls to use `https` which I seems preferable.

PR #722, opened 2018-11-28
feat: cleanup repository before starting up IPFS

hacdias:
If `api` file exists and it is the same as the api address defined in the `config` file, it is removed.

If `api` file exists and it is different from the one in the `config` file, the user will be asked if he wants to remove it or not.

Nothing os this will happen if there is a `repo.lock`.

License: MIT


olizilla:
@lidel Do we have to prompt the user if we can tell it's from an unclean shutdown? I guess it's safer at this point to draw attention to it rather than hide it.

lidel:
@olizilla I think prompt with two options (`ipfs repo fsck` or quit) is both safe and removes surprises.

It creates an opportunity to backup `$IPFS_PATH`  before running `ipfs repo fsck` or quitting without doing anything and figuring out what went wrong with `<insert super custom setup user created to own demise>`  :)

hacdias:
I can do the dialog as @lidel mentioned, it won't require many changes. It would be nice if `js-ipfs-api` supported that command. I'll need to run it directly.

hacdias:
@olizilla done: https://github.com/ipfs/js-ipfs-http-client/issues/901

lidel:
Keep in mind we are talking about cleaning up lockfiles BEFORE go-ipfs starts, so http API can't be used anyway and running CLI command directly (with env `$IPFS_PATH`) is the only way to do it :)

hacdias:
@olizilla @lidel so... the coded solution works, but I don't like it though. But it works. Could you try?

src/utils/daemon.js
```diff
@@ -1,7 +1,12 @@
 import IPFSFactory from 'ipfsd-ctl'
 import logger from './logger'
+import { join } from 'path'
+import { spawnSync } from 'child_process'
+import { dialog } from 'electron'
```
olizilla: This file now mixes multiple concerns. Let's split out the external process finding and calling and the electron dialog elements into small functions that we can test and compose.

src/utils/daemon.js
```diff
@@ -1,7 +1,12 @@
...
 
-export default async function createDaemon (opts) {
+import findExecutable from 'ipfsd-ctl/src/utils/find-ipfs-executable'
```
olizilla: neat trick!

src/utils/daemon.js
```diff
@@ -61,3 +66,48 @@ export default async function createDaemon (opts) {
...
+
+  return true
+}
+
+export default async function (opts) {
+  try {
+    const ipfsd = await createDaemon(opts)
+    return ipfsd
+  } catch (e) {
+    if (!e.message.includes('ECONNREFUSED')) {
```
olizilla: I don't think looking for just `ECONNREFUSED' gives us enough info. I think we want to do the work of checking if there is an `api` file and no repo lock file as well, otherwise this could be a case of not being able to connect to an api for a daemon that _is_ running. Offering to `fsck` the repo in that case would be unhelpful. @lidel what do you think?

I still think we should run `repo fsck` automatically in the common case wherer there is no lockfile and the api file points to the same location as the API address in the config file.

src/utils/daemon.js
```diff
@@ -61,3 +66,48 @@ export default async function createDaemon (opts) {
 
   return ipfsd
 }
+
+async function cleanup (opts) {
+  const option = dialog.showMessageBox({
+    type: 'warning',
+    title: 'IPFS Desktop',
+    message: `We coudln't start because IPFS didn't shutdown correctly last time. We can try to
```
olizilla: There is a typo in coudln't and, let's refer to IPFS Desktop rather than `We`

```suggestion
    message: `IPFS Desktop found an api file in ${IPFS_HOME}, but was unabled to connect to the address ${apiAddr}`. Would you like to try running `ipfs repo fsck` to remove the file and try again?
```

src/utils/daemon.js
lidel: Running `repo fsck` to fix `ECONNREFUSED` makes sense only if:
- `api` is present 
  - AND points at the same multiaddr as `config`  
    - AND (`$IPFS_PATH/repo.lock` OR `$IPFS_PATH/datastore/LOCK`) is present

hacdias: @lidel if is present and points to the **same** address? I thought that in that case it was safe to run it without consent of the user. I was thinking we should ask them only if it was different.

hacdias:
@lidel I'm aware. I've a commit I didn't push yet, waiting for more feedback so we can close this soon.

src/utils/daemon.js
lidel: yes, sorry, I meant "not the same", fixed my comment :)

src/utils/errors.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  app.exit(1)
+}
+
+export function fsckError (path, addr) {
+  const option = dialog.showMessageBox({
+    type: 'warning',
+    title: 'IPFS Desktop',
+    message: `IPFS Desktop found an api file in ${path}, but was unabled to connect to the address ${addr}. Would you like to try running 'ipfs repo fsck' to remove the file and try again?`,
```
lidel:
```suggestion
    message: `IPFS Desktop found an 'api' file in ${path}, but was unable to connect to the address ${addr}. Would you like to try running 'ipfs repo fsck' to remove the file and try again?`,
```

src/utils/daemon.js
```diff
@@ -1,5 +1,62 @@
...
+  const apiFile = join(path, 'api')
+
+  if (!await fs.pathExists(apiFile)) {
+    return true
+  }
+
+  const cfgFile = join(path, 'config')
+  const cfg = await fs.readJSON(cfgFile)
+  const cfgAddr = cfg.Addresses ? cfg.Addresses.API : ''
+  const addr = (await fs.readFile(apiFile)).toString()
```
lidel: File often ends with newline character because people create it manually with `echo "/ip4/127.0.0.1/tcp/5001" > ~/.ipfs/api`  :)

```suggestion
  const addr = (await fs.readFile(apiFile)).toString().trim()
```

src/utils/daemon.js
```diff
@@ -1,5 +1,62 @@
...
+
+async function configure (ipfsd) {
+  let origins = []
+  try {
+    origins = await ipfsd.api.config.get('API.HTTPHeaders.Access-Control-Allow-Origin')
+  } catch (e) {
+    logger.warn(e)
+  }
+
+  if (!origins.includes('webui://-')) origins.push('webui://-')
```
olizilla: I wonder how this could be abused. A fake scheme is a neat trick, but I don't know enough to reason about how it could be exploited if we added it to the list of allowed origins. I guess a site would have have been loaded from the webui:// scheme. What happens if a web-ext is allowed to registered itself as handling that scheme?

What are our other options here?

src/utils/errors.js
```diff
@@ -0,0 +1,54 @@
...
+- **Chrome Version**: ${process.versions.chrome}
+
+**Error**
+
+\`\`\`
+${e.stack}
+\`\`\`
+`
+
+export function genericError (e) {
```
olizilla:
```suggestion
export function showErrorMessage (e) {
```
In general, function names should say what the **do**. more verby than nouny.

src/utils/errors.js
```diff
@@ -0,0 +1,54 @@
...
+  if (option === 1) {
+    shell.openExternal(`https://github.com/ipfs-shipyard/ipfs-desktop/issues/new?body=${encodeURI(issueTemplate(e))}`)
+  } else if (option === 2) {
+    app.relaunch()
+  }
+
+  app.exit(1)
+}
+
+export function fsckError (path, addr) {
```
olizilla:
```suggestion
export function showRepoApiFileErrorMessage (path, addr) {
```

src/utils/errors.js
```diff
@@ -0,0 +1,54 @@
+import { app, dialog, shell } from 'electron'
+
+const issueTemplate = (e) => `Please describe what you were doing when this error happened.
```
olizilla: This file is nice and tidy. Much clear now it's split out from the createDaemon logic.

src/utils/daemon.js
```diff
@@ -46,18 +107,6 @@ export default async function createDaemon (opts) {
...
-    logger.warn(e)
-  }
-
-  if (!origins.includes('webui://-')) origins.push('webui://-')
-  if (!origins.includes('https://webui.ipfs.io')) origins.push('https://webui.ipfs.io')
-
-  await ipfsd.api.config.set('API.HTTPHeaders.Access-Control-Allow-Origin', origins)
-  await ipfsd.api.config.set('API.HTTPHeaders.Access-Control-Allow-Methods', ['PUT', 'GET', 'POST'])
-
+  await configure(ipfsd)
```
olizilla: I'd assume we'd need to update the config before starting the repo? or at least we'd need to restart the deamon if we do update the config.

src/index.js
```diff
@@ -61,43 +30,30 @@ async function setupConnection () {
...
   try {
     // Initial context object
     let ctx = {
-      ipfsd: await setupConnection()
+      ipfsd: await createDaemon(config)
+    }
+
+    // createDaemon has changed the config object,
+    // but didn't add the repo variable.
+    if (updateCfg) {
```
olizilla: Let's simplify this. `createDaemon` shouldn't mutate the config that is passed to it. Let's pull all the setting of defaults up into this file, so we don't have to guess about when we might need to update the stored config.

src/utils/daemon.js
hacdias: This is an interesting question. We could create a local server instead. There aren't many options. Running the file directly (through `file://`) doesn't work (first thing I tried) and running a local server will also obligate us to add a 'safe' origin to the list.

We could also use `webui.ipfs.io` but that would require internet connection to work, wouldn't it? Or we could fetch the latest version using the user's node.

src/index.js
hacdias: @olizilla then we have to test if there is an IPFS_PATH env variable and if not, use the home path. All of that is already done by ipfsd-ctl which will get updated in case of something changes. I just didn't want to repeat the code.

src/utils/daemon.js
hacdias: Yes, I'll change that 😄

src/index.js
olizilla: That's totally reasonable. How would you re-arrange things so that createDaemon doesn't mutate the options object that is passed to it? How can we make the code clearer that we may provide a path, but if we dont, we want to ask the ipfsd what it's path is?

hacdias: Simple: we could initially put the config settings as sth like:

```json
{
	"type": "go",
	"path": "",
	"flags": [...defaultFlags]
}
```

`createDaemon` would not replace the settings with the defaults and after getting the `ipfsd` object we could do:

```javascript
if (cfg.path === '') {
	cfg.path = ipfsd.repoPath
	// update cfg file...
}
```

hacdias: @olizilla just updated the code with a clearer version and this way `createDaemon` doesn't change the opts object.

src/index.js
```diff
@@ -81,23 +22,33 @@ async function run () {
...
     // Initial context object
     let ctx = {
-      ipfsd: await setupConnection()
+      ipfsd: await createDaemon(config)
+    }
+
+    /// Update the path if it was blank previously.
+    // This way we use the default path when it is
+    // not set.
+    if (config.path === '') {
```
olizilla: Good job @hacdias ✨ 💎 ✨  this is so much clearer now!

src/utils/daemon.js
olizilla: My main concern is about updating the users config for an external daemon. @lidel what do you think?

src/utils/store.js
```diff
@@ -4,8 +4,15 @@ const store = new Store()
...
   store.clear()
-  store.set('seenWelcome', false)
-  store.set('config', null)
+
+  // default config
+  store.set('config', {
+    type: 'go',
+    path: '',
+    flags: ['--migrate=true', '--routing=dhtclient'],
+    keysize: 4096
```
olizilla:
```suggestion
    keysize: 2048
```

https://github.com/ipfs/go-ipfs/blob/481136b10585e0d98bc9a5ec76953f644b43232b/cmd/ipfs/init.go#L25

https://github.com/ipfs/go-ipfs/issues/1107

out of curiosity where did the 4096 value come from?

src/utils/store.js
```diff
@@ -4,8 +4,15 @@ const store = new Store()
 
 if (store.get('version') !== 4) {
   store.clear()
-  store.set('seenWelcome', false)
-  store.set('config', null)
+
+  // default config
+  store.set('config', {
```
olizilla:
```suggestion
  store.set('ipfsConfig', {
```

`config` is a really generic name for a property in a config store!

src/utils/store.js
hacdias: @olizilla honestly no idea 😆

src/utils/daemon.js
hacdias: Btw, as you said, to update the config we need to do it before starting the daemon. And to change the config before starting the daemon, we need to edit the `config` file directly as we're doing now. So it won't affect remote nodes

src/utils/daemon.js
```diff
@@ -1,42 +1,109 @@
...
+      IPFS_PATH: path
+    }
+  })
+}
+
+async function cleanup (path) {
+  const apiFile = join(path, 'api')
+
+  if (!await fs.pathExists(apiFile)) {
+    return true
```
olizilla: Let's make the api for `cleanup` clearer. Having `cleanup` return true if it doens nothing is suprising. I think we need more info than just a boolean return value.

There are few other issues
- `const cfg = await fs.readJSON(cfgFile)` could throw if the config isn't valid json. This function should be robust and try it's best to detect what's going on.
- `if (addr === cfgAddr && (lockFile || datastoreLock)) {` I think the situatuon we're looking for is just "there is an api file and no lockFile and no datastroreLock.", then it's safe to just run fsck. If there is a lockfile and an apiFile, but we failed to connect to the api, it'd be ok to ask the user if we should try and fsck. We shouldn't offer to fsck the repo if there is alredy a daemon running that we just can't connect to for some reason.

Let's pull this function out to a helper `detectOrphanApiFile` that returns true if the api File is present and the same as the config default value, and that there is no repo.lock and no datastore/LOCK.

olizilla: > - `api` and `repo.lock` are present but no daemon running AND  `api` points at the same multiaddr as `config`  

@lidel is it possible to assert that no daemon is running? is an  `ECONNREFUSED` from the api enough?

src/utils/daemon.js
lidel: Actually.. we have big control over the browser in IPFS Desktop, just like we do in IPFS Companion.  This means we should not need to set the CORS header in daemon. 

Electron (Chromium engine) is adding `Origin: null` to fetch requests made from `webui://` pages which acts as a sort-of-false-positive trigger for CORS validation, resulting in `403 Forbidden`:


 



**Note:** IPFS Companion works around this by removing Origin header from requests sent to the API from js-ipfs-http-client running in its background page:

- https://github.com/ipfs-shipyard/ipfs-companion/blob/v2.6.2/add-on/src/lib/ipfs-request.js#L107-L144

I believe if  we add similar fix to IPFS Desktop the need for `origins.push(...)` will disappear.

hacdias: @lidel I'll try that out! Thank you

src/utils/daemon.js
lidel: I'd say `ECONNREFUSED` its the best platform-agnostic hint we can get.

Alternative is to execute platform-specific process enumeration(`tasklist` on windows, `ps -A` on linux, `ps something` on mac) to see if process is running, which may get blocked and not work reliably in more locked down systems.

src/utils/daemon.js
lidel: Let's keep this PR only about `ipfs repo fsck`
@olizilla @hacdias 
I extracted CORS to separate issue: https://github.com/ipfs-shipyard/ipfs-desktop/issues/728

olizilla:
Thanks @Stebalien. We're going to keep it simple for this pass. If ipfsd-ctl tries to connect to an existing api and fails, and there is an api file present, then we'll offer the user something like:

failed to start dialog

on the assumption that users who are setting api files directly will have a clue as to what is going on, while users that just have an unclean repo state have an option to `fsck` it, which is a command they may not have been aware of, and it may fix it for them.

src/utils/daemon.js
```diff
@@ -1,63 +1,104 @@
...
   } catch (e) {
-    logger.warn(e)
-  }
+    if (!e.message.includes('ECONNREFUSED')) {
+      throw e
+    }
 
-  if (!origins.includes('webui://-')) origins.push('webui://-')
-  if (!origins.includes('https://webui.ipfs.io')) origins.push('https://webui.ipfs.io')
+    if (!showRepoApiFileErrorMessage(ipfsd.repoPath, ipfsd.apiAddr)) {
```
olizilla: This is neat, but it's now assuming that there is an `api` file rather than checking for it. Either we need to check for it so we can assert it in the error message, or we need to update the wording of the error message to something more like:

"IPFS Desktop failed to connect to an existing ipfs api at ${ipfsd.apiAddr}. This can happen if you run `ipfs daemon` manually and it is not shutdown cleanly. Would you like to try running `ipfs repo fsck` to remove the lock and api files from ${ipfsd.repoPath} and try again?

src/utils/daemon.js
```diff
@@ -1,63 +1,104 @@
...
+  if (!origins.includes('webui://-')) origins.push('webui://-')
+  if (!origins.includes('https://webui.ipfs.io')) origins.push('https://webui.ipfs.io')
+
+  cfg.API.HTTPHeaders['Access-Control-Allow-Origin'] = origins
+  cfg.API.HTTPHeaders['Access-Control-Allow-Methods'] = ['PUT', 'GET', 'POST']
+
+  await fs.writeJSON(cfgFile, cfg)
+}
+
+async function spawn ({ type, path, keysize = 0 }) {
```
olizilla: What happens if `keysize` is set to 0, and why set a default for it here?

src/utils/daemon.js
```diff
@@ -1,63 +1,104 @@
...
-        resolve()
-      })
+      resolve()
     })
+  })
+}
+
+export default async function (opts) {
+  const ipfsd = await spawn(opts)
+  await configure(ipfsd)
```
olizilla: If ipfsd connects to an existing daemon process, and we change it's config, we would need to restart that process for the config to be applied.

hacdias: @olizilla we're now editing the config file directly so we won't even be able to change the config of remote daemons. Perhaps it would be better to revert this change and use the API.

src/utils/daemon.js
hacdias: `keysize` is only used when initializing a repo. We don't actually need to set it to 0 here. The default is set on store.js when we add the config if it's not present.

src/utils/daemon.js
hacdias: Just updated the error message.

src/utils/daemon.js
hacdias: Ping @olizilla

hacdias: @olizilla I'm merging this so we can speed up dev. This piece of code will be removed on #736 so let's not take too much time here.

PR #730, opened 2018-12-05
feat: add spectron startup test

olizilla:
This PR adds a basic (the text book) spectron test, modified just enough to make it work with our app

Right now our app won't close after testing, and I have narrowed it down to the way that we keep the webui window active in
https://github.com/ipfs-shipyard/ipfs-desktop/blob/f06cc5919682ba80009bceb07c39a210dfd63e66/src/hooks/webui/index.js#L29-L33

This seems to cause the call to `app.stop()` in the spectron api to fail. To fix this and more importantly to be kind on users system resources, i think we should avoid keeping the webui window running. I think we should open it when the app is first run, and destroy it when the user closes the window, and create a new one if they choose to open it again. This will require refactoring how we handle the config toggle, as it expects to be able to pull the window from the context. I think we should decouple that.

TODO:

- [ ] **DO LATER** Refactor webui to create and destroy windows. Refactor things that depend on finding a stable webuiWindow property on the context to not do that.
- [x] Assert more interesting things about the windows rather than just "do the exist"
- [x] Tests depend on the babel'd output, so we should either stop using features that require babel in our app code, or run babel before running the tests, and and a watch mode.


hacdias:
@olizilla I pushed a small fix to master and then merged it onto this branch. It fixes the issue where the app wouldn't close on `app.stop`.

src/index.js
```diff
@@ -25,10 +25,17 @@ async function run () {
   let config = store.get('ipfsConfig')
 
   try {
+    const ipfsd = await setupConnection()
+
+    app.on('quit', () => {
+      ipfsd.stop(err => {
```
hacdias: This is not sync so we are never sure this code will actually be executed. What if the program exits before the callback is called?

src/menubar/index.js
```diff
@@ -38,16 +38,6 @@ export default async function (ctx) {
     }
 
     ipcMain.on('app.quit', async () => {
-      logger.info('Stopping daemon')
-
-      await new Promise((resolve, reject) => {
```
hacdias: That's why I did did it in an async but sync way here. Am I right or...?

src/index.js
hacdias: Another interesting information is:

> Note: On Windows, this event will not be emitted if the app is closed due to a shutdown/restart of the system or a user logout.

So we will basically show the fsck error every time the users logs in. I'll check tomorrow if there are any events on Windows for shutdown/restart/logout.

olizilla: Good point! The reason for the change was to pull the shutdown logic out of the menubar code and up next to the ipfsd creation so you can see where it is created and stopped in the same place. I'm assuming that there are OS specific ways to quit the app that don't involved hitting the power button on the menubar.

olizilla: If I open the webui window and hit `⌘ q` to quit the app on osx, it shuts down the app. So we need to try and stop the ipfsd in in a global quit handler if possible.

hacdias:
So, let's take a look of what's missing. By injecting `IPFS_PATH` in the environment, we may be able to try out this startup scenarios:

- No previous repository;
- Existent okay repository;
- Existent not-okay (with `api`) file repository;
- Make tests work with Babel;

@olizilla anything missing?

hacdias:
@olizilla @fsdiogo I created some tests but they only seem to be passing on macOS. They work locally on Windows. I followed the instructions for Spectron on [Travis and AppVeyor](https://github.com/electron/spectron/#continuous-integration) but it still doesn't seem to work. I will AFK for the next couple of hours. It would be awesome if you could take a look at this please.

hacdias:
Just an update: I just updated the tests and they now also mock the home directory so they can be safely run at your computer and they won't change your configurations, nor your IPFS Desktop settings.

On Windows, it seems to be opening tons of empty command line windows and it makes the tests hang. I have to manually click 'Quit' on IPFS Desktop to finish each test manually. Here's a related issue on Spectron: https://github.com/electron/spectron/issues/60.

Despite this issue, they seem to be testing what they should test. I'm waiting for the CI to see how they behave on other OSes. Although I'm expecting them to fail on Linux (as they were before) and on Windows (due to the command line windows...). Hopefully macOS tests will pass, as they were before.

hacdias:
One more update: I noticed the reason for tests to be failing is not the daemon, is not spectron. It's the Web UI window. Most of the times it can't quit so I decided to remove the Window completely and now the tests pass... This might also be related to the issue @lidel has reported which required multiple clicks on 'Quit' to close IPFS Desktop (#781).

In other "news", I decided to allow failures on Linux because of two reasons: only OS where the tests don't pass **and** we are setting Linux support as experimental since it doesn't work well on all desktop environments: https://github.com/ipfs-shipyard/ipfs-desktop/issues/788#issuecomment-458978056.

hacdias:
I managed to fix this in the meanwhile - and actually I don't think it's related to #781 anymore - by setting the node integrations to `true` on testing environment. I don't yet understand why is it needed but hey! The tests pass.

Ref.: https://github.com/electron/spectron/issues/347

hacdias:
@olizilla @fsdiogo @lidel could you take a look at this please? Perhaps we could merge this!

PR #736, opened 2018-12-07
wip: remove the need for CORS headers

### LINKED ISSUE #728: Remove the need for setting CORS headers
> Issue extracted from https://github.com/ipfs-shipyard/ipfs-desktop/pull/722#pullrequestreview-181190975

## Summary

Right now IPFS Desktop is automatically adding CORS headers to `$IPFS_PATH/config`:

https://github.com/ipfs-shipyard/ipfs-desktop/blob/8f912ce816e2cd2e1c634fe8cdab54e2eb37cedf/src/utils/daemon.js#L59-L64

Without the above, Web UI won't load.
We should look into ways we can make it work without the need for touching `API.HTTPHeaders`, eg. see how IPFS Companion deals with the same problem. 

## Background Discussion

@olizilla asked:
> I wonder how this could be abused. A fake scheme is a neat trick, but I don't know enough to reason about how it could be exploited if we added it to the list of allowed origins. I guess a site would have have been loaded from the webui:// scheme. What happens if a web-ext is allowed to registered itself as handling that scheme?
> 
> What are our other options here?
> My main concern is about updating the users config for an external daemon. @lidel what do you think?

@lidel suggested:

> Actually.. we have a lot of control over the browser (request headers etc) in IPFS Desktop, just like we do in IPFS Companion.  This means we should not need to set the CORS header in daemon. 
> 
> Electron (Chromium engine) is adding `Origin: null` to fetch requests made from `webui://` and internal contexts to HTTP API which acts as a sort-of-false-positive trigger for CORS validation, resulting in `403 Forbidden`:
> 
> 
>  
> 
> 
> 
> **Note:** IPFS Companion works around this by removing Origin header from requests sent to the API from js-ipfs-http-client running in its background page:
> 
> - https://github.com/ipfs-shipyard/ipfs-companion/blob/v2.6.2/add-on/src/lib/ipfs-request.js#L107-L144
> 
> This makes those requests  "no longer CORS" in view of go-ipfs. 
> I believe if  we add similar fix to IPFS Desktop the need for `origins.push(...)` will disappear. 
> 
>

hacdias:
Please see #728.

Needs:
- [ ] https://github.com/ipfs-shipyard/ipfs-webui/pull/899


hacdias:
@lidel @olizilla we will need to enable `window.ipfs` on Web UI or create another way (like `window.ipfsDesktop` so Web UI knows we're talking to Desktop.

IMO, it would be easier to just enable window.ipfs when calling the ipfsReduxBundle on Web UI. It would just work. Right now, I'm injecting the `ipfs` object into the window. It seems to be working. It just isn't recognized by the Web UI

hacdias:
@olizilla @lidel I just updated this PR to use a version from Web UI that **uses** `window.ipfs`. You can try it out. We don't need CORS headers anymore.

We just need to merge https://github.com/ipfs-shipyard/ipfs-webui/pull/899 and this one 😄 

Suggestions?

hacdias:
will tidy up this PR.

PR #741, opened 2018-12-11
feat: improve security of electron app

hacdias:
Prevent any URL that is not from our app to open inside one of our windows. This takes into account the recommendations of [Electron's team](https://github.com/electron/electron/blob/master/docs/tutorial/security.md).

**Update:** content security policy headers seem not to work when web security is set to false.

### DIFF
--- a/src/hooks/app-events.js
+++ b/src/hooks/app-events.js
@@ -1,0 +1,19 @@
+import { app, shell } from 'electron'
+
+export default function () {
+  app.on('web-contents-created', (_, contents) => {
+    contents.on('will-navigate', (event, url) => {
+      const parsedUrl = new URL(url)
+
+      if (parsedUrl.origin !== 'webui://-') {
+        event.preventDefault()
+        shell.openExternal(url)
+      }
+    })
+
+    contents.on('new-window', (event, url) => {
+      event.preventDefault()
+      shell.openExternal(url)
+    })
+  })
+}
--- a/src/hooks/app-events.js
+++ b/src/hooks/app-events.js
@@ -21,7 +21,7 @@ export default function () {
     callback({ // eslint-disable-line
       responseHeaders: {
         ...details.responseHeaders,
-        'Content-Security-Policy': ['default-src \'none\'']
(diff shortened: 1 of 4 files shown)


src/hooks/app-events.js
```diff
@@ -0,0 +1,28 @@
...
+      event.preventDefault()
+      shell.openExternal(url)
+    })
+  })
+
+  session.defaultSession.webRequest.onHeadersReceived((details, callback) => {
+    callback({ // eslint-disable-line
+      responseHeaders: {
+        ...details.responseHeaders,
+        'Content-Security-Policy': ['default-src \'none\'']
```
olizilla: How did you come to pick this as the CSP? I just ran the [CSP Labratory plugin](https://addons.mozilla.org/en-US/firefox/addon/laboratory-by-mozilla/) on Web UI and it came up with

```
 default-src 'none'; connect-src 'self' http://127.0.0.1:5001/api/v0/config/show http://127.0.0.1:5001/api/v0/get http://127.0.0.1:5001/api/v0/id http://127.0.0.1:5001/api/v0/object/get http://127.0.0.1:5001/api/v0/stats/bw http://127.0.0.1:5001/api/v0/swarm/peers; font-src 'self'; img-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline' 
```

which i think we can trim down, but it's a start.

hacdias: @olizilla maybe it could be trimmed down to:

```
default-src 'none'; connect-src 'self'; font-src 'self'; img-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline' 
```

I set it to `default-src none` because it was the less permissive of all and it worked. Which now leads me to think that disabling web security might disable CSP. I couldn't find any info about it. But it is in fact weird that setting the default-src to none doesn't change anything. I'll investigate.

hacdias:
@olizilla just updated it here.

olizilla:
were you able to verify if it is applied when webSecurity is false?

PR #742, opened 2018-12-12
feat: add auto updater

hacdias:
This is the bare minimum code needed to set up auto updating. It is supposed to automatically detect the repository URL through `package.json`. It looks on the releases for the binaries and `latest.yaml` files produced by `electron-builder`.

We can listen to events and modify some behavior if needed.

License: MIT

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1554,14 +1554,12 @@
     "bluebird": {
       "version": "3.5.3",
       "resolved": "https://registry.npmjs.org/bluebird/-/bluebird-3.5.3.tgz",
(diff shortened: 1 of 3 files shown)


olizilla:
@hacdias what would need to happen to make that work?

hacdias:
@olizilla Travis for Windows isn't working well with secrets so we should use Jenkins. That's just it: define the 'GH_TOKEN' environment variable and change add a small release script on package.json telling to release on new tags.

hacdias:
@olizilla I'll merge this in the meanwhile because that issue doesn't block merging this.

PR #744, opened 2018-12-12
feat: create draft release on tag

hacdias:
Adds the required info to auto-upload the binaries to GitHub. By default, the created releases are supposed to be only `drafts`. We can then review them and publish.

We still need to add the required env variable to Jenkins.

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "build:webui:minimize": "shx rm -rf src/hooks/webui/app/static/js/*.map && shx rm -rf src/hooks/webui/app/static/css/*.map",
     "build:icons": "svgr -d src/icons node_modules/ipfs-css/icons && standard --fix src/icons/*.js",
     "build:babel": "babel src --ignore src/hooks/webui/app --out-dir out --copy-files",
-    "build:binaries": "electron-builder -p never"
+    "build:binaries": "electron-builder -p onTag"
   },
   "pre-commit": [
     "lint"
@@ -57,7 +57,8 @@
     "nsis": {
       "include": "build/nsis.nsh",
       "oneClick": false
-    }
+    },
+    "publish": ["github"]
   },
   "devDependencies": {
     "@babel/cli": "^7.2.0",


olizilla:
>  By default, the created releases are supposed to be only drafts.

Can you explain this a little more please?

hacdias:
@olizilla it creates a draft on GitHub releases so then we can really publish. Drafts are hidden and only visible to us.

lidel:
@hacdias just to be sure I got the proper context, we are following the flow linked below or something else?
-  https://www.electron.build/configuration/publish#recommended-github-releases-workflow

hacdias:
@lidel that will just happen if we set to `onTagOrDraft`. But that's basically it. We could even do that, to always have the latest binaries in a drafted release!

olizilla:
Just to echo what @lidel said, can we check we are on the same page for what we do next? The [docs](https://www.electron.build/configuration/publish#recommended-github-releases-workflow) say

> Recommended GitHub Releases Workflow
> 1.  Draft a new release. Set the “Tag version” to the value of version in your application package.json, and prefix it with v. “Release title” can be anything you want. For example, if your application package.json version is 1.0, your draft’s “Tag version” would be v1.0.
> 2. Push some commits. Every CI build will update the artifacts attached to this draft.
> 3. Once you are done, publish the release. GitHub will tag the latest commit for you.
>
> The benefit of this workflow is that it allows you to always have the latest artifacts, and the release can be published once it is ready.

**and** earlier on 
> CLI --publish option values:

onTag | on tag push only
-- | --
onTagOrDraft | on tag push or if draft release exists
always | always publish
never | never publish

> But please consider using automatic rules instead of explicitly specifying publish:
>    If CI server detected, — onTagOrDraft.
>    If CI server reports that tag was pushed, — onTag.
> Release will be drafted (if doesn’t already exist) and artifacts published only if tag was pushed.

olizilla:
We should set up channels for releases! cc @lidel (he loves a good beta channel)
- https://www.electron.build/tutorials/release-using-channels

hacdias:
@olizilla the current Travis setup builds successfully but Travis on Windows still doesn't work with secrets so it won't be able to upload.

olizilla:
So we can either

- Publish osx and linux installers from travis, and you add windows ones manually
**or**
- Try the suggested `travis.yaml` from the docs, which builds the windows version on linux

hacdias:
Closing in favour of #702.

PR #754, opened 2018-12-18
fix: cleanup before starting (fixes #746)

### LINKED ISSUE #746: ipfsd not stopped when quitting app from OS
Yesterday I installed a binary with the latest version from master, on Windows, and I set it to startup on login. Since IPFS Desktop doesn't shutdown correctly when Windows turns off, I always get this message:

image

I didn't found, yet, a way to listen for Windows shutdown but there must be a way because I know that many apps can even prevent Windows from turning off.

hacdias:
If there is a `config` file, we run `ipfs repo fsck` command to cleanup the lock files. If there is no `config` file, we expect a remote daemon so we do nothing.

`repo fsck` isn't supposed to remove the locks if a daemon is already running so there is no issue (tested on Windows).

### DIFF
--- a/src/utils/daemon.js
+++ b/src/utils/daemon.js
@@ -16,6 +16,29 @@ function repoFsck (path) {
   })
 }
 
+function startStop (path) {
+  logger.info('Starting daemon to clean up the locks')
+  const opts = {
+    env: {
+      ...process.env,
+      IPFS_PATH: path
+    },
+    timeout: 10000,
+    killSignal: 'SIGINT',
+    stdio: 'inherit'
+  }
+  let out = ''
+  try {
+    const exec = findExecutable('go', app.getAppPath())
+    out = execFileSync(exec, ['daemon'], opts)
+    console.log(out)
+    
+  } catch (e) {
+    console.log(out)
+    if (!e.message.includes('ETIMEDOUT')) throw e
+  }
+}
+
 async function spawn ({ type, path, keysize }) {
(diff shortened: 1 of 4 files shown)


src/utils/daemon.js
```diff
@@ -1,11 +1,19 @@
...
 import { app } from 'electron'
 import { execFileSync } from 'child_process'
 import findExecutable from 'ipfsd-ctl/src/utils/find-ipfs-executable'
 
-function repoFsck (path) {
+async function cleanup (path) {
+  logger.info(`Entering cleanup stage`)
+
+  if (!await fs.pathExists(join(path, 'config'))) {
+    logger.info(`'config' file does not exist. Aborting cleanup`)
```
olizilla: We should show a custom error message here, as the default one asks the user if they want to report the error, and we don't want that. We want to explain that the api file and lack of config file mean we can't start the app.

hacdias: Why? In this case there is no `config` file, so ipfsd-ctl will connect to a remote node. Unless we want to prevent Desktop from connecting to a remote node. Is that so?

olizilla:
> I didn't add the code only if we got ECONNREFUSED because this way we prevent it already and simplify the code.

I missed that point! We should only clean the repo if we hit a problem. Please do wait for an ECONNREFUSED err before trying to fsk.

hacdias:
@olizilla done!

PR #768, opened 2018-12-22
German (de.json)

JayBrown:
German translations

### DIFF
--- a/src/locales/de.json
+++ b/src/locales/de.json
@@ -1,0 +1,9 @@
+{
+  "turnOff": "IPFS Desktop beenden",
+  "ipfsNode": "IPFS-Node",
+  "status": "Status",
+  "files": "Dateien",
+  "explore": "Erkunden",
+  "peers": "Peers",
+  "settings": "Einstellungen"
+}


hacdias:
Hello @JayBrown! Thanks a lot for this. The translations are automatically synced from [Transifex](https://www.transifex.com/ipfs/ipfs-desktop/translate/#de). Could you head to this [link](https://www.transifex.com/ipfs/ipfs-desktop/translate/#de) and fill the translations there? Otherwise I can merge it and add them myself. 😄 Whatever is best for you.

hacdias:
I see you already added it to Transifex! Thanks 😄

hacdias:
Uh, I believe that should be a translator's choice. But I'll "cc" @lidel here as he is more experienced with the translation project 😄

JayBrown:
Well, it's a community effort, so if the community says it should be translated as *Knoten*, that's how it's going to be. The status yesterday, however, was a mix of both *Node* and *Knoten* across all IPFS projects, so I changed everything to *Node*, but I'm not against changing all to *Knoten*, even if it *is* a really ugly word. ;)

lidel:
@JayBrown Thank you for cleaning this up. In general translator have a lot of freedom, the goal is to make things look and sound good in particular language, it does not need to be 1:1 translation. 

Not sure how useful this is, but there is a way to create "preferred" translations for specific nouns, verbs etc.  You add a word or a phrase as a term to _Glossary_:
> 

then open Glossary and provide preferred German translation there:

> 

Items present in Glossary get underlined to indicate they require additional focus during translation and a suggested translation will be present on the right  pane (this will be visible to all future translators and strings). 

Ref: https://docs.transifex.com/setup/glossary/adding-terms

JayBrown:
Yeah, I did that, but then quickly noticed that you need to create a glossary for *every* app, and there doesn't seem to be an ipfs master glossary. (Don't even know if that's possible.)

lidel:
I have admin on Transifex and found a way to define a global Glossary (shared by all IPFS projects), so I did some cleanup and assigned all projects to use shared one:

> 

Hope this helps in keeping terms aligned across projects :)

Issue #765, opened 2018-12-21
Improve UX for "auto add screenshot" 


momack2:
I love the auto-add screenshot function - but I think it could use some UX smoothing:
- When taking a screenshot, send a notification that users can click to see their screenshot in the desktop UI
- Right clicking a file in the desktop ui should also pop open it's option menu (to share, download, etc)
- Have a step between keyboard shortcut and selecting what to screenshot (full desktop, active window, choose selection) - otherwise I might accidentally screenshot and start providing something confidential! Right now if you have two windows it screenshots both - which is super unexpected IMHO
- I get a 524 (or worse, an infinite loading screen) when trying to "share" and view a screenshot I just took on the ipfs or cloudflare gateways -- I think this might be a wider gateway issue - but FYI this makes that feature unusable (@eefahy)

I bet these are probably shared desktop/webUI items - but found them in this context so figured I'd surface them here. =] Great work! Looking forward to becoming a regular user of this feature

hacdias:
Hey @momack2! Glad you liked it and brought some feedback. I'm splitting this issue into two (this one here and https://github.com/ipfs-shipyard/ipfs-webui/issues/917 for Web UI's issues).

> When taking a screenshot, send a notification that users can click to see their screenshot in the desktop UI

Yup, that can be done and I that was what was initially planned when this featured was introduced (like one year ago). But at the time there was a bug that didn't send notifications on Windows, but that is solved now.

> Have a step between keyboard shortcut and selecting what to screenshot (full desktop, active window, choose selection) - otherwise I might accidentally screenshot and start providing something confidential!

I know we an screenshot the whole screen and Electron's window. I wonder if we can get a good screenshot of windows of other apps. I will investigate.

> I get a 524 (or worse, an infinite loading screen) when trying to "share" and view a screenshot I just took on the ipfs or cloudflare gateways

I believe this is #750.

---

We'll work into these. Thank you 😄 Any other ideas?

/cc @ipfs-shipyard/gui

PR #775, opened 2018-12-26
feat: improve screenshots ux

hacdias:
With this PR, when adding a screenshot with more than one monitor, a folder will be created with the screenshots from each screen instead of appending the screen name to the end of the file name. This way it is easier to share.

We also add notifications for when the screenshot was successfully created and if you click on it, it will open the screenshot on the Web UI.

On error, it also shows a notification.

This addresses some of the issues on #765.

It also starts using `share.ipfs.io` instead of the main gateway link.


TheChiefMeat:
Sorry for the delayed response; Christmas ate my time away.

Seems good to me; I get the folder with both screenshots inside.

PR #779, opened 2019-01-03
feat: quit with context menu

### LINKED ISSUE #764: Menubar icon disappears after 'turning off' IPFS
I would have expected that there is a greyed-out menubar item that I could go to to fire up IPFS again, especially since the icon used in the interface is a "power button". Not sure if this is a bug, or expected behavior at the moment.

hacdias:
This fixes #764 by using the toggler in the menubar itself only for IPFS and adding a context menu that can be used for three things: quit the app, quick access to settings and logs directory, which might be useful.

ex

This works well on Windows. Please test on macOS and Linux!!

### DIFF
--- a/src/lib/register-daemon.js
+++ b/src/lib/register-daemon.js
@@ -3,27 +3,41 @@ import { app } from 'electron'
 
 export default async function (ctx) {
   let config = store.get('ipfsConfig')
-  ctx.ipfsd = await createDaemon(config)
+  let ipfsd = null
 
-  // Update the path if it was blank previously.
-  // This way we use the default path when it is
-  // not set.
-  if (config.path === '') {
-    config.path = ctx.ipfsd.repoPath
-    store.set('ipfsConfig', config)
+  ctx.getIpfsd = () => ipfsd
+
+  ctx.startIpfs = async () => {
+    ipfsd = await createDaemon(config)
+
+    // Update the path if it was blank previously.
+    // This way we use the default path when it is
+    // not set.
+    if (config.path === '') {
+      config.path = ipfsd.repoPath
+      store.set('ipfsConfig', config)
+    }
   }
+
+  ctx.stopIpfs = async () => {
+    return new Promise((resolve, reject) => {
+      ipfsd.stop(err => {
+        if (err) {
+          return reject(err)
+        }
+
+        ipfsd = null
+        resolve()
+      })
+    })
+  }
+
+  await ctx.startIpfs()
 
   app.once('will-quit', async (event) => {
     event.preventDefault()
     logger.info('Stopping daemon')
-
-    await new Promise((resolve, reject) => {
-      ctx.ipfsd.stop(err => {
-        if (err) reject(err)
-        else resolve()
-      })
-    })
-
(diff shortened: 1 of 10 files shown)


fsdiogo:
I still haven't looked at the code  but this doesn't seem to be working on macOS.

Left-clicking the icon is showing the menu that should appear when right-clicking:

mac

hacdias:
Ugh. On Windows it works... I remember we tried doing this on the previous version but that "bug" was happening on macOS and Linux. I hoped it was fixed already though. Probably it's the expected behaviour.

hacdias:
It also seems that right-click event is only supported on Windows and macOS: https://electronjs.org/docs/api/tray#event-right-click-macos-windows

@olizilla @fsdiogo @lidel what do you suggest?

hacdias:
Just tried a different approach and use the right click event to show up the context menu:

https://github.com/ipfs-shipyard/ipfs-desktop/blob/b3c9aaab29707e27ee834fe35bc2e26a1e8cb0a3/src/lib/menubar/index.js#L45-L48

This **won't work on Linux** but I hope it fixes for macOS. On Windows it keeps working.

fsdiogo:
Working on macOS.

hacdias:
@lidel which are the guidelines or usual ways of quitting a menubar app on Linux?

hacdias:
@lidel that's rad! Btw, only macOS needs the right-click event handling. On Windows it works with the context menu too.

No, you shouldn't need to click twice. Did you wait some time before trying again? Maybe it was trying to quit IPFS. We can try to debug that.

lidel:
@hacdias I think the issue may be that I use remote API:
- It quits instantly when i use embedded go-ipfs, but..
-  I need to click `Quit` twice to exit when I use remote one  (`$IPFS_PATH/api` is present)

olizilla:
Can confirm, works as intended for me too on macOS

src/lib/register-daemon.js
```diff
@@ -1,29 +1,51 @@
...
+        ipfsd = null
+        resolve()
       })
     })
+  }
+
+  await ctx.startIpfs()
 
+  app.once('will-quit', async (event) => {
+    event.preventDefault()
```
lidel: @hacdias Fix from https://github.com/ipfs-shipyard/ipfs-desktop/pull/779/commits/88b586cf0e84b22b3b3b0a196848e4428ff915ad did not help on linux, however if I remove this line `Quit` action starts working as expected  (with both embedded and remote nodes):

```suggestion
```


I think we were just breaking shutdown flow:
> Calling `event.preventDefault()` will prevent the default behaviour, which is terminating the application.
> https://github.com/electron/electron/blob/master/docs/api/app.md#event-will-quit

Is  `event.preventDefault()` necessary? My intuition is that shutdown will happen after `quit` event which happens after `will-quit` anyway.

hacdias: @lidel, yes, it is needed otherwise ipfsd.stop might never be called. If you notice, that's a **once** listener and in the end we call **app.quit** again so there shouldn't be any problems unless the app.quit is never reached.

PR #798, opened 2019-01-22
feat: add context menu 'Add to IPFS' on Windows

hacdias:
Click on Add to IPFS on a context menu and a file will be sent to IPFS and a notification

image

- If Desktop is not running, it will start and add it.
- If Desktop is started with daemon off, nothing will be done.
- Caveat: if multiple files are selected it is the same as calling the command for each one of them, which might result in tons on notifications if the user selects lots of files. Doubt it will happen a lot of times since it's easier to send a directory. We should discuss on this.


hacdias:
Perhaps we should add to user config an option to disable the context menu. What do you think @ipfs-shipyard/gui? Enabled by default though.

olizilla:
I'm against adding config options unless there is significant demand for them. What is the use-case where you wouldn't want it?

hacdias:
@olizilla at least on Windows all programs allow you to choose whether or not you want it. That was the only reason

hacdias:
@olizilla although if we opt for not having that option, the code I wrote can also be simplified down to installer instructions instead of code adding the ctx menu.

hacdias:
Maybe we could just not add that now and see if anyone asks for that.

hacdias:
@olizilla I decided to use your suggestion and so I cleaned up the code a bit.

hacdias:
What I want you to try out, if installing this on Windows isn't an option, is the following command:

```
npm start -- --add="/path/to/file/or/dir"
```

And see if it works.

hacdias:
Ping @olizilla @lidel @fsdiogo

Issue #807, opened 2019-01-29
Handle ipfs://, ipns:// and dweb: system-wide


hacdias:
We have kind of a bit stale issue (#35) about handling `ipfs://` at a system wide level. With the help of Electron Builder ([docs](https://www.electron.build/generated/platformspecificbuildoptions)) we can easily configure one or more schemes to be handled by our program. Ideas ideas!!

- Handle `ipfs://` URLs and show a dialog asking whether to open it in the gateway or explore in IPLD Explorer. Wouldn't a 'Add to my repository' option be interesting too?
Or we could have a way of opening it inside Web UI and then Web UI would have those options (instead of showing a dialog).
- Shall we handle `ipns://` to and redirect to the browser?
- Shall we also support `.ipfs` files that can contain an IPFS path? https://github.com/ipfs-shipyard/ipfs-companion/issues/641

---

- macOS: works.
- Windows: works.
- Linux: ??

/cc @ipfs-shipyard/gui

hacdias:
Sure, let's keep `.ipfs` file support on a different PR! 😄 In the meanwhile, I'll open a PR about the scheme handler.

hacdias:
@ipfs-shipyard/gui working well on Windows. On macOS is working if Desktop is already running. Didn't test on Linux but it is not supposed to work yet.

Ref.: https://github.com/electron-userland/electron-builder/issues/1413

hacdias:
Update: working on macOS.

sneak:
question - if ipfs-desktop is running a local ipfs daemon (with gateway), shouldn't it be rewriting ipfs: and ipns: URLs to localhost:8080/{ipfs,ipns} instead of the public gateway?

lidel:
AFAIK it is by design, to make links always work (even when your local node is shut down).
If you have [ipfs-companion](https://github.com/ipfs/ipfs-companion/) browser extension it will redirect to your local node and public gateway won't get hit.

hacdias:
Yeah, we did it that way to make sure they were always available. _But_ we can detect if the daemon is on, and if so, use the local gateway.

PR #808, opened 2019-01-29
feat: handle ipfs://, ipns:// and dweb: protocols

hacdias:
Ref.: #807

- Implements `ipfs://`, `ipns://`, `dweb:/ipfs/` and `dweb:/ipns/` protocol handlers.
- It required some more configuration on Windows since Electron Builder doesn't automatically add the handler configuration to NSIS script (see https://github.com/electron-userland/electron-builder/issues/623).
- On macOS and Linux it is supposed to work. Please try out.

### DIFF
--- a/src/lib/second-instance.js
+++ b/src/lib/second-instance.js
@@ -2,6 +2,16 @@ import { app } from 'electron'
 import { extname, basename } from 'path'
 import { logger, i18n, notify, notifyError } from '../utils'
 
+function getFile (argv) {
+  for (const arg of argv) {
+    if (arg.startsWith('--add')) {
+      return arg.slice(6)
+    }
+  }
+
+  return ''
+}
+
 async function copyFile (launch, ipfs, hash, name, folder = false) {
   let i = 0
   const ext = extname(name)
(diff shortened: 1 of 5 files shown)


src/lib/second-instance.js
```diff
@@ -81,9 +66,27 @@ const addToIpfs = ({ getIpfsd, launchWebUI }) => async (_, argv) => {
...
+  const handler = (_, argv) => {
+    for (const arg of argv) {
+      if (arg.startsWith('--add')) {
+        return addToIpfs(ctx, arg.slice(6))
+      } else if (arg.startsWith('ipfs://')) {
+        return openLink('ipfs', arg.slice(7))
+      } else if (arg.startsWith('ipns://')) {
+        return openLink('ipns', arg.slice(7))
+      } else if (arg.startsWith('dweb:/ipfs/')) {
+        return openLink('ipns', arg.slice(11))
```
lidel: :eyes: 
```suggestion
        return openLink('ipfs', arg.slice(11))
```

hacdias: Thanks!

fsdiogo:
If IPFS Desktop is not running on macOS, the first time we try to access `ipfs://<hash>` it only gets the app running with no redirect.

Other than that it seems to be working on macOS.

PR #809, opened 2019-01-29
feat: improve add to IPFS handler

hacdias:
Refactors the add handler (for context menus) and added a feature for macOS users: drop files into the icon of the app (not the tray) to add them. It probably won't work if the app is not yet running because the event is fired before starting IPFS.

License: MIT


fsdiogo:
> a feature for macOS users: drop files into the icon of the app (not the tray) to add them.

Do you mean the dock icon? It's not accepting dropped files. The tray icon (menu bar) accepts dropping files but then they are not added to the files page.

## Dock

> dock

## Tray (menu bar)

> menu

hacdias:
macOS `open-file` event removed due to https://github.com/electron/electron/issues/14029

PR #810, opened 2019-01-30
feat: drop files in the tray (macOS)

hacdias:
License: MIT

### DIFF
--- a/src/lib/add-to-ipfs.js
+++ b/src/lib/add-to-ipfs.js
@@ -81,6 +81,13 @@ export default async function (ctx) {
     handleArgv(argv)
   })
 
+  // macOS tray drop files
+  ctx.menubar.tray.on('drop-files', async (_, files) => {
+    for (const file of files) {
+      await addToIpfs(ctx, file)
+    }
+  })
+
   // Checks current proccess
   if (process.env.NODE_ENV !== 'development') {
     await handleArgv(process.argv)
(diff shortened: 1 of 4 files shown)


fsdiogo:
Not working. I can take a look later because coding in the dark is no fun (❍ᴥ❍Ʋ)

fsdiogo:
So, I don't know what happened the last time that I tested this, but **it is working**. Sorry for my misdirection 😪

There is just one detail that may have put me off last time: when the app window is open and you're in the files page, after adding the file through the tray icon everything works but the files list doesn't get updated, so you don't see the added file. You have to manually change tabs or refresh the list to see it.

A solution to this if to check if the window is opened and we're in the files page, and if so, force a refresh. Is this easy peasy?

hacdias:
@fsdiogo just added a reload after adding the files. Please check if it works, please 😄

fsdiogo:
It works @hacdias 👍

I don't want to nitpick, but reloading the whole Web UI seems a bit harsh to me, specially if you have it open in another page.

Can we check if the current open page is the files page, and if so just fetch the files list? If we're on another page we don't need to refresh the Web UI, because when we switch to the files page the files list will be fetched to show the current state.

fsdiogo:
If that's too much of a hassle we can merge this now and take care of it later too.

hacdias:
We could also just move to `/files` page. After all, after adding the files it would be nice to have the Files page ready. No?

hacdias:
@fsdiogo I've changed so it updates to `/files` page. We can improve this a little bit later.

Issue #665, opened 2018-10-05
Tweak default config of IPFS node


lidel:
Default settings of go-ipfs are not optimized for Desktop use.

Let's collect things IPFS Desktop should set during repo init.

Initial ideas:
-  [x] Enable GC - https://github.com/ipfs-shipyard/ipfs-desktop/pull/826
-  [x] We should limit on the number of connections, as it is known to choke cheap consumer routers and wifi cards. Some insight on sane  defaults is max of ~300 conns:  https://github.com/ipfs/go-ipfs/issues/1482#issuecomment-427346155 - https://github.com/ipfs-shipyard/ipfs-desktop/pull/828
- [x] Make sure dhtclient remains to be set :) (https://github.com/ipfs-shipyard/ipfs-desktop/issues/619)

cc #661

olizilla:
See also https://github.com/ipfs/go-ipfs/issues/4989

lidel:
I think now that https://github.com/ipfs-shipyard/ipfs-desktop/pull/828 is merged we got MVP  :)

PR #828, opened 2019-02-17
feat: default minimum and maximum connections

hacdias:
Set the range to 200-300. The default was 300-600. Ref: #665

License: MIT

### DIFF
--- a/src/utils/daemon.js
+++ b/src/utils/daemon.js
@@ -45,6 +45,22 @@ async function spawn ({ type, path, keysize }) {
         keysize: keysize
       }, e => {
         if (e) return reject(e)
+
+        try {
+          // Set default mininum and maximum of connections to mantain
+          // by default. This only applies to repositories created by
+          // IPFS Desktop. Existing ones shall remain intact.
+          const configFile = join(ipfsd.repoPath, 'config')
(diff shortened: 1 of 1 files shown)


olizilla:
I'm in favour of this change, but for posterity, we should replace it with a desktop profile in IPFS core when it is available https://github.com/ipfs/go-ipfs/issues/4989

olizilla:
@lidel good news, the logic only runs on init!

https://github.com/ipfs-shipyard/ipfs-desktop/blob/62317e02bb6f534c5806c0c6c204bfacf09999e7/src/utils/daemon.js#L39-L41

olizilla:
From @whyrusleeping notes it sounds like we should set the `ConnMgr` values to

```json
"LowWater": 50,
"HighWater": 200,
"GracePeriod": "300s"
```

Otherwise
- `Discovery.MDNS.enabled is `true` by default
- `DisableNatPortMap` is `false` by default

but we could explicitly set them here. I'm not sure what we'd need to change to ensure

> relay addresses should be advertised

hacdias:
Merging this now with the defaults decided here. Let's hope a desktop profile is created soon so we can just use it! 😄

PR #833, opened 2019-02-20
feat: version indicators

### LINKED ISSUE #832: Show desktop version and ipfs version
Could we add

- desktop version number
- ipfs version number

to the context menu? So it's easy to find out what versions of things you have.

hacdias:
Closes #832 by adding Desktop and IPFS version to the top of the context menu.

image

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "electron-updater": "^4.0.6",
     "filesize": "^4.1.2",
     "fs-extra": "^7.0.1",
-    "go-ipfs-dep": "^0.4.18",
+    "go-ipfs-dep": "0.4.18",
     "i18next": "^15.0.4",
     "i18next-electron-language-detector": "0.0.10",
     "i18next-icu": "^1.1.2",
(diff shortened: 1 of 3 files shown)


hacdias:
@olizilla could you take a look to see how it looks on macOS? If the options can't be grayed out, we can add a click action

olizilla:
Grays out nicely

screenshot 2019-02-20 at 11 31 23

But what if we had

- ipfs-desktop v0.7.0.rc.1
- go-ipfs v0.4.18

as links to the releases on github?

olizilla:
Could add a grayed out  section header "Versions"

hacdias:
image

Done, can you check out if it works?

olizilla:
Looks good, but doesn't open the release page on macos. Also it looks like the code will open the desktop releases page for both items.

hacdias:
@olizilla fixed!

package.json
```diff
@@ -118,7 +118,7 @@
     "electron-updater": "^4.0.6",
     "filesize": "^4.1.2",
     "fs-extra": "^7.0.1",
-    "go-ipfs-dep": "^0.4.18",
+    "go-ipfs-dep": "0.4.18",
```
olizilla: Good idea. We should be very deliberate about when we update the go-ipfs dep.

PR #866, opened 2019-03-12
feat: just a menubar

### LINKED ISSUE #732: Remove React from the menubar
> @olizilla also, the menubar is so simple, I'm not sure we need react. it'd load faster if we had a static html file and some click handlers

As Oli said, probably the menubar would load faster without React. We used it initially because we thought we were going to have much more features embedded in the menubar itself.

### LINKED ISSUE #857: Visualise the status of the ipfs daemon and be robust to it not stopping.
If you leave the IPFS daemon running for a long time, trying to stop it can take a long time. The user has no clue what is going on, so it just seems like the app is stalled... I took a screen recording of a particularly suprising session, that stalled, then crashed, then installed an update

https://ipfs.io/ipfs/QmWEDikLypHaWYNJC3ftmGNkuZCogeSoMSnwAp7UaLyCdz/ipfs-desktop-no-node-status.mp4

The app should not crash if the daemon takes a long time to stop. We may have to consider killing the process or offering the user a way to do it, if it does take a really long time. Command line users are just another ctrl+c away from killing it.

We could indicate the current status of the daemon by adding a label to the "Status" entry on the menubar with labels for ["stopped", "starting...", "connected", "stopping..."] or similar. I've used "connected" for the all ok label, as that mirrors the language in the web ui.

We would be good if the power button visually indicated if the "power" was on or not, as companion does. It may make more sense as a slider toggle switch, as per https://github.com/ipfs-shipyard/ipfs-companion/issues/689#issuecomment-469496633

### LINKED ISSUE #864: Upgrade to go-ipfs 0.4.19
Folks want the latest shiny. See https://github.com/ipfs-shipyard/ipfs-webui/issues/988

We need to unblock the [0.7 release](https://github.com/ipfs-shipyard/ipfs-desktop/issues/788) so we can get it to them!

### LINKED ISSUE #865: Move to menubar menu only
Most menubar apps have the same behaviour when using the right or left click and right now we have different actions for both. On left click, we open the real React menubar and with the right click we option a native menu with some options.

Since this goes against what most do, probably people would get confused and we thought about moving it only to a menubar menu app. With this we could simplify the user experience (on all OSes) and reduce lots of dependencies we need right now.

/cc @ipfs-shipyard/gui

hacdias:
I've taken a bit of time today to work on this since I really enjoyed the idea - and I liked the result very much. Don't get mistaken by the 'wip' in the title. The only thing missing are changes to the README 😄The software is workiiiiiing!

It is working on macOS and Windows. It is supposed to work better on Linux now since we don't have the issue of which 'screen' is showing (some envs would just show the menubar with options on either click).

Windows

macOS

Please read #865 for further background information. @ericronne suggested to collapse the Web UI options (from Status to Peers) into one 'Open Web UI'. Although I actually like the current configuration. Please try it out!

> Wonder if in the future we wouldn't be able to ditch Electron and open Web UI on user's browser and use Qt for the menubar.

- Closes #732: we remove React completely from the app.
- Closes #857: we now have a beautiful status information with colors!
- Closes #864: we updated the dependencies and go-ipfs!!
- Closes #865: well... that is what this is about.

## TODO

- [ ] Update README screenshot


hacdias:
@ipfs-shipyard/gui I need y'all here 😄

hacdias:
@lidel 

> Not sure why left-click does not work, perhaps it is i3bar I use for tray.

I think what I just pushed might fix that.

> Settings should be moved out of "Advanced" to the top level, and be in the same order as in Web UI (under "Peers"), to minimize user confusion :)

Yas. Talking about that: we only have three options in settings related to Desktop. Perhaps they could be checkbox-type buttons on a Preferences sub-menu where the user could activate them and deactivate them.

> "Check for updates" could be moved to "Advanced" submenu to declutter UI a bit

Yeah!

> Overall, I like it a lot! This change makes it feel light and snappy.

Feels light, but it isn't light... In the future I'd like to investigate possible ways of moving out of Electron. It would be lighter and perhaps even/actually more native.

olizilla:
I really like it. I think we could drop the third divider.

something like

```
- IPFS is running
- About IPFS Desktop
---
- View Status
- View Files
- View Settings
---
Advanced ▸           Stop IPFS daemon
Quit                 Open repo directory
                     Open logs directory
                     Check for updates
                     _Versions_
                     IPFS Desktop 0.7-rc4
                     go-ipfs 0.4.19                              
```

- We don't need to list out all peers and explore from the main navbar, just the most useful ones. Also this avoids the "am i opening the Desktop app, or the Web UI, what should I call it?" question.
- Move "Stop IPFS" to the advanved menu. It should be available, but not one of the main things.
- At this early stage, while we're still getting the auto update mechanism working, it's useful to have an easy way to see the versions, so push them into the advanced section too.

And what about adding a link to docs.ipfs.io? this'd  draw more attention to it and encourage us to add a section on Desktop and Web UI.

olizilla:
also of note, the status link isn't working for me.

hacdias:
@olizilla will restructure that way! Seems a good idea.

> also of note, the status link isn't working for me.

Neither for me, thanks for noticing.

hacdias:
image

fsdiogo:
> Move "Stop IPFS" to the advanced menu. It should be available, but not one of the main things.

@olizilla I don't think so, especially when you want to start the node again: 

desktop

It's pretty odd finding that function buried, and having the `About` being the second item too. The `About` should be between `Advanced` and `Quit` (most of the mac apps have between before the quit).

hacdias:
> @fsdiogo: It's pretty odd finding that function buried, and having the About being the second item too. The About should be between Advanced and Quit (most of the mac apps have between before the quit).

Yeah, I agree with those two.

hacdias:
And perhaps we could hide the versions behind 'About IPFS Desktop'. Instead of opening the repository, we could show a simple informational dialog.

fsdiogo:
Hmm yeah, if that dialog is customisable we can have the links to the repos there!

olizilla:
@fsdiogo I was following the example of Docker for Mac. I don't find it weird, but I don't feel strongly about it.

I don't think "Stop IPFS" should be in the main menu. A user has to restart IPFS to make a config change, but we should offer that option in the UI, when they make a config change. In the meantime we could add a "Restart IPFS" option to either the advanced, or the main menu if you feel strongly that it should be there.

My assumption is that for most users, we'd like them to be able to set there node up once (ideally just run with the defaults) and then leave IPFS running. If IPFS takes too much resources, we should limit it. We just dropped the number of open connections it'll try and make from 600+ to about 150, which should help.

hacdias:
@lidel how do you feel about this?

fsdiogo:
> I don't think "Stop IPFS" should be in the main menu. A user has to restart IPFS to make a config change, but we should offer that option in the UI, when they make a config change. In the meantime we could add a "Restart IPFS" option to either the advanced, or the main menu if you feel strongly that it should be there.

@olizilla the idea of restarting through Web UI is great 👌 I feel strongly that it shouldn't be in the advanced, and we should have the option to `Start` and `Stop` instead of just restarting.

docker

@hacdias as we're talking about Docker, this `About` window looks good don't you think?

info

lidel:
- Start/Stop/Restart actions could land under the Status item.
- Web UI items also could be moved it own submenu, but personally I prefer to keep them at the top-level. I'd like "Files" to be super easy to access.
- I don't like "View" and "Open" being repeated in UI, so I would just drop them from labels.

```
IPFS is running ▸    Restart
                     Stop
---
Status
Files
Peers
Settings  
---
Advanced ▸           Logs Directory
                     Repo Directory
                     IPFS Desktop Configuration File

About ▸              About IPFS Desktop
                     Check for Updates
                     _Versions_
                     ipfs-desktop 0.7-rc4
                     go-ipfs 0.4.19
Quit
```

hacdias:
@fsdiogo I really like Docker's about page. Although remember: that's a native window. We can't do one as native as that one. What we could do is create a very simple web page to show. Again - it would be really nice if in the feature we could have really native apps for each platform.

---

@lidel I like the idea of starting/stopping and restart being under the status item.

I agree with dropping 'Open' and 'View' labels.

I personally don't like having those options under 'About'. It feels weird to me.

---

Another interesting thing is that on macOS we usually see 'Per Word Capitalisation' and on Windows we have 'Per sentence capitalisation'.

olizilla:
I think we are zooming in on an ideal list. I like @lidel's proposal, pulling the start and restart under the status seems neat.

In general think the menu items should be things you can **do**, so _verbs_ rather than _nouns_. 

- I feel strongly that actions that open directories should start with "Open" e.g " Open logs directory"
- I can live with the links to the web ui sections not starting with "View", as I agree, the reptition looks odd, but in general, I'm in favour of explicit and clear over terse and more visually pleasing where menu options are concerned.
- I like @fsdiogo's ask for an about dialog, but perhaps lets keep it as menu options for now to keep this PR focused.

```
IPFS is running ▸    Restart
                     Stop
---
Status
Files
Settings  
---
Advanced ▸           Open logs directory
                     Open respository directory
                     Open configuration file

About ▸              _Versions_
                     ipfs-desktop 0.7-rc4
                     go-ipfs 0.4.19
                     ----
                     Check for Updates...
                     View on Github

Quit
```

hacdias:
@olizilla I agree with that disposition. What do you think @lidel @fsdiogo?

hacdias:
Done:

image

README.md
```diff
@@ -17,11 +17,37 @@ IPFS Desktop allows you to run your IPFS Node on your machine without having to
...
 - [Install](#install)
 - [Contribute](#contribute)
     - [Translations](#translations)
 - [FAQ](#faq)
 
+## Features
+
+### IPFS daemon always running
+
+IPFS Desktop's main feature is to allow you to have the IPFS daemon always running in the background. But fear not! If you need to stop it, you can do it under the 'Advanced' options.
```
olizilla: ❤️

src/icons/status/gen_status.sh
```diff
@@ -0,0 +1,26 @@
+#/usr/bin/env bash
+
+png () {
+  middle=$(($3 / 2))
+  center=$((middle - 1))
+
+  convert -size "$3x$3" xc:transparent \
```
olizilla: very cool

olizilla: not for this PR, but would be good to document the dependencies for running this script.

src/lib/tray.js
```diff
@@ -0,0 +1,166 @@
...
+    menu = buildMenu(ctx)
+    tray.setContextMenu(menu)
+    tray.setToolTip('IPFS Desktop')
+    updateStatus(status)
+  }
+
+  const updateStatus = data => {
+    status = data
+
+    menu.getMenuItemById('ipfsIsStarting').visible = status.starting && !status.done
```
olizilla: How about just setting the status to one of these string values in the first place? with the collection of booleans, it's harder to be sure that we have covered every eventuality. If status can only be 1 string, then we can quickly scan the deamon init code to see all the possible values,  and we can just use it directly here.

src/icons/status/gen_status.sh
hacdias: Will do later :) I can even just add a comment there or add to the README. There's something I don't really like about that: not sure if the sizes are perfect. They look good but it seems that they're not as correct as they should be. But we can take a look at that later.

src/lib/tray.js
hacdias: Yes, I wanted to make those constants and just compare. I'll do that in another PR.

src/lib/tray.js
```diff
@@ -0,0 +1,166 @@
...
+          label: i18n.t('openLogsDir'),
+          click: () => { shell.openItem(app.getPath('userData')) }
+        },
+        {
+          label: i18n.t('openRepoDir'),
+          click: () => { shell.openItem(store.get('ipfsConfig.path')) }
+        },
+        {
+          label: i18n.t('openConfigFile'),
+          click: () => { shell.openItem(store.path) }
```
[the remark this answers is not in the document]
hacdias: Yeah, we already had that possibility. It might be useful for someone who wants to change the flags that we use to run the daemon.

hacdias:
@olizilla I'll update the circle in another PR and the other things you've mentioned!

alanshaw:
w00t! ship it!

src/lib/tray.js
hacdias: it would be even better to just change the text and the icon... But this is not native and Electron doesn't let us do that.

lidel:
> Not sure why left-click does not work, perhaps it is i3bar I use for tray.
(this should not block this PR, we can investigate this later).

For the record:
I just re-tested in [v0.7.0](https://github.com/ipfs-shipyard/ipfs-desktop/releases/tag/v0.7.0) and the issue is fixed there (menu pops up for both left and right clicks) :rocket:  :+1:

PR #876, opened 2019-03-17
💄: prettier DMGs

hacdias:
Then:

Now:


hacdias:
I'm merging this. We can't change the font color (it isn't perfectly readable), but it looks much better. We can change anything later.

PR #878, opened 2019-03-18
docs: make downloads the main thing

olizilla:
Add a download section at the top, and add info on package mangers to the install section.


hacdias:
Cool! Thanks!

PR #894, opened 2019-04-02
docs(README): optimize hero image

lidel:
PNG was 777 KB and slowed down README load.
MozJPEG (http://squoosh.app) trimmed it down to 35KB :)


hacdias:
Thanks!

PR #896, opened 2019-04-05
WIP: Add `ipfs` to `PATH`

hacdias:


hacdias:
@lidel @olizilla need your feedback about what we currently do: I don't feel great about the idea of doing nothing when the user has IPFS already, but I think we have margin to play since we use a script and don't link directly to the binary.

Could you please review the scripts to see if they're OK?

hacdias:
WIP update on Linux: only working on deb.

hacdias:
As per https://github.com/electron-userland/electron-builder/issues/3436 it seems we can't run install scripts on some platforms. Ideally we could do everything on runtime. @lidel is there a way on Linux to ask permission temporarily so we can modify /usr/local/bin?

hacdias:
@lidel yes, there were some issues on macOS, but they're fixed now 😄

@olizilla would you mind to try building and installing?

electron-builder.yml
```diff
@@ -3,6 +3,12 @@ appId: io.ipfs.desktop
 directories:
   buildResources: assets/build
 
+extraFiles:
```
hacdias: @lidel Probably we could use `extraResources` here, no? https://www.electron.build/configuration/contents#extraresources

Just to make sure it goes to `resources` on Linux and `Resources` on macOS.

hacdias: I've updated this on 7c5f6bba63c534f2e34f5d4f2b1fed2c031e7ad9. @lidel could you please make sure it still works on Linux?

electron-builder.yml
```diff
@@ -35,6 +41,15 @@ linux:
...
+rpm:
+  fpm:
+    - "--after-install=assets/build/after-install.sh"
+    - "--after-remove=assets/build/after-remove.sh"
+
+deb:
+  afterInstall: assets/build/after-install.sh
+  afterRemove: assets/build/after-remove.sh
+
 snap:
```
olizilla: How does it work for `snap` installs?

src/early/ipfs-script.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  // Ignore during development because the paths are not the same.
+  if (process.env.NODE_ENV === 'development') {
+    logger.info('[ipfs on path] unavailable during development')
+    return
+  }
+
+  // Don't make any changes if IPFS already exists...
+  if (fs.existsSync('/usr/local/bin/ipfs')) {
```
olizilla: `install.sh` from a go-ipfs distribution will try to install in either `/usr/local/bin` or `/usr/bin` https://github.com/ipfs/go-ipfs/blob/622dbf7f266e513c6de00cddc5cb851aa5cbfe61/cmd/ipfs/dist/install.sh#L9

perhaps we should try and exec `ipfs` here to see if it exists on the users path.

src/early/ipfs-script.js
```diff
@@ -0,0 +1,30 @@
...
+    logger.info('[ipfs on path] was not added, already exists')
```
olizilla: For a subsequent PR, we could prompt the user here. 

"you appear to have a version of the `ipfs` command line tool already. Would you like to let ipfs-desktop replace it with the a version that it can keep up to date?"

...that kind of thing.

electron-builder.yml
hacdias: As mentioned above, it doesn't. Only on `deb` for now.

src/early/ipfs-script.js
hacdias: Yes, that's the goal. I'm just making this one as simple as possible.

electron-builder.yml
lidel: Tested, works fine :)

assets/bin/ipfs.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+resources=$(dirname "$(dirname "$(test -L "$0" && readlink "$0" || echo "$0")")")
```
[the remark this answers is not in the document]
lidel: Added in  7e8c171

olizilla:
woo haa! It works!

```console
$ cat `which ipfs`
#!/usr/bin/env bash

# Get the full path of "resources/" directory (resolving the symlink if needed)
resources=$(dirname "$(dirname "$(test -L "$0" && readlink "$0" || echo "$0")")")
# Get the full path to ipfs binary bundled with ipfs-desktop
ipfs="$resources/app.asar.unpacked/node_modules/go-ipfs-dep/go-ipfs/ipfs"

exec "$ipfs" "$@"

$ ipfs version
ipfs version 0.4.19
```

Issue #727, opened 2018-12-05
Provide `ipfs` command line tools on install


olizilla:
Atom.app provides the `apm` command when it is installed, as do many other apps. It would be rad if installing IPFS Desktop added the `ipfs` command to your PATH if it's not already present.

We already bundle the binaries, so it's just a matter of figuring out how to make them available.

olizilla:
Can you tackle this next... it'd really help us with the "easy install" story for https://github.com/ipfs/website/pull/299

lidel:
Some notes from IRC about `PATH` on Linux:

#### Symlinking to `/usr/local/bin` is possible only during install

After install ipfs-desktop runs under non-root user and cant write to `/usr/local/`

```console
$ ln -s ./ipfs /usr/local/bin/ipfs
ln: failed to create symbolic link '/usr/local/bin/ipfs': Permission denied
```

That being said, we should be fine with symlinking to `/usr/local/bin/ipfs` during install on Linux. Symlink may conflict with plain `go-ipfs` package, but if desktop user wants to run a different version of go-ipfs than one shipped with ipfs-desktop, they can do it either by  `docker run --rm -it --net=host ipfs/go-ipfs:v0.4.18` or pointing at remote node via `~/.ipfs/api`


#### Adding `ipfs` binary to `PATH` needs to account for  shell quirks 

This is alternative to symlinking which would enable us to add/remove `ipfs` from the `PATH` of current user via a checkbox in UI, but comes with a lot of edge cases:

> The main difference with shell config files is that some are only read by "login" shells (eg. when you login from another host, or login at the text console of a local unix machine). these are the ones called, say, `.login` or `.profile` or `.zlogin` (depending on which shell you're using).
> 
> Then you have config files that are read by "interactive" shells (as in, ones connected to a terminal (or pseudo-terminal in the case of, say, a terminal emulator running under a windowing system). these are the ones with names like `.bashrc`, `.tcshrc`, `.zshrc`, etc.
> 
> `bash` complicates this in that `.bashrc` is only read by a shell that's both interactive and non-login, so you'll find most people end up telling their `.bash_profile` to also read `.bashrc` 
> 
> – https://stackoverflow.com/a/415444

hacdias:
[...]

PR #897, opened 2019-04-06
feat: add 'ipfs' to PATH on Windows

hacdias:
During installation, append `ipfs-desktop/assets/bin-win` to the `PATH` on Windows. It's also removed during installation. The directory contains a script that executes our bin script that redirects the arguments to the actual binary. That way we have some flexibility about what we can do in the future.

The code used to manipulate the `PATH` comes from this resources:

- [Path Manipulation](https://nsis.sourceforge.io/Path_Manipulation)
- [Smartools Installer](https://www.smartmontools.org/browser/trunk/smartmontools/os_win32/installer.nsi?rev=4110#L636)

Also, I had to disable "warnings as errors" because Electron Builder is using Macros for custom install and uninstall sections so when building the installer, NSIS didn't know if we were calling the functions in the correct places. It threw a warning that was treated as an error. Not anymore.

Addresses #727.


djdv:
I encountered an unrelated issue with `npx` during install (https://github.com/zkat/npx/issues/209)
But besides that, everything worked as expected on my machine. :+1:
After installation, the PATH var was modified to include the mentioned path, when uninstalled, it was no longer there in new environments.
Running `ipfs.cmd` gave me no issues after installation.

hacdias:
Per @djdv feedback, I'm merging this PR 😄 Thanks for checking it out.

PR #905, opened 2019-04-16
fix: move i18n to late hooks

hacdias:
While building other PR, I noticed the translations weren't available at early stages of startup and that's because i18n returns a promise. Here I change the mechanism so we wait for the translations to be ready.

License: MIT

### DIFF
--- a/src/late/auto-updater.js
+++ b/src/late/auto-updater.js
@@ -1,5 +1,6 @@
 import { autoUpdater } from 'electron-updater'
-import { logger, i18n, notify, quitAndInstall } from '../utils'
+import i18n from 'i18next'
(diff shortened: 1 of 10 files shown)


olizilla:
Could you explain a little more about the early and late hooks? I missed the PR wher that was introduced and I dont have context on why it was needed.

hacdias:
@olizilla the name 'early' and 'late' hooks are basically just names. The early ones are things that must run as soon as the application boots (such as registering protocol handlers to avoid missing calls) and the 'late' ones are what we had on lib before and that don't need to be run ASAP.

Also, early ones must be sync, while late ones can be async.

Issue #880, opened 2019-03-19
Release 0.8.0 🏯 'ipfs' on PATH and bug fixing


hacdias:
After the huge release that 0.7.0 was (see #788), we need to start thinking about 0.8.0 😄 Here's the todo:

### What's left to be done

- [x] Add `ipfs` command line tools to the `PATH` (#727, https://github.com/ipfs-shipyard/ipfs-webui/pull/1016).
- [x] Figure out why we needed 'System Events' on macOS (#753).
- [x] Don't shutdown if there was an issue `fsck`ing the repository (#899, #835, #837).
- [x] Support macOS dark theme on native dialogs (#918)
- [x] Make dialogs more consistent and cohesive (#919) 
- [x] Avoid auto-update unnecessary errors (#892)
- [x] Update to Electron 5 (#917)
- [x] Allow settings when the daemon is offline (https://github.com/ipfs-shipyard/ipfs-webui/pull/1023)
- [x] Work with bandwidth metrics disabled (https://github.com/ipfs-shipyard/ipfs-webui/pull/1024)
- [x] Add folder via file-input fails in electron (#884)
- [x] Update Web UI hash
- [x] Update to `go-ipfs` v0.4.21.

### Postponed

- [ ] Hijack native macOS screenshot shortcut #879 
- [ ] Use a random port when the defaults are being used (#912)
- [ ] Investigate on how to make boot faster through [Electron Link](https://www.npmjs.com/package/electron-link)
- [ ] #785 preview PDFs

### After the release

- [x] Update Homebrew
- [x] Update Chocolatey

---

Please remember that this list list may change 🐉

olizilla:
rather than a roadmap, lets update this issue to give it a pet name, and define just what is going in, like we did in https://github.com/ipfs-shipyard/ipfs-desktop/issues/788

hacdias:
@olizilla could you check this issues, please:

- https://github.com/ipfs-shipyard/ipfs-webui/pull/1016
- https://github.com/ipfs-shipyard/ipfs-webui/pull/1023
- https://github.com/ipfs-shipyard/ipfs-webui/pull/1024

Also, how's the upload/1MB issue?

olizilla:
[...]

PR #906, opened 2019-04-16
feat: ipfs on PATH on runtime

### LINKED ISSUE #727: Provide `ipfs` command line tools on install

hacdias:
Closes #727.

See https://github.com/ipfs-shipyard/ipfs-desktop/pull/906#issuecomment-485549688


hacdias:
Just noticed that if the user says 'No', every time they start IPFS Desktop, they'll be prompted.

What do you think? If we add an option 'No, and don't ask' we should provide a way to undo it if the user happens to change its mind.

Also, keep in mind this only affects macOS users.

fsdiogo:
Being nagged every time a user starts Desktop is no good.

I'd say we should just ask once, but have an option (a button for example) in the desktop specific settings to trigger the replacement.

hacdias:
Hmm... I'm not sure if the status block would be a good way because this only affects macOS and it could require more work than what it's worth.

What if we show once and, if the user says no (or if /usr/local/bin/ipfs gets manually changed), we add an option (perhaps under Advanced or even in the main menu) to add it to PATH.

What do you think?

cwaring:
That sounds like a good compromise to the this implemented and we can always revisit later.

If you install IPFS Desktop and IPFS isn't in the users path does we automatically set it up?

I ask because presenting people with decisions on initial load can be hostile, especially if it feels permanent and the user doesn't yet understand the benefits.

My suggestion would be:
- a) Add the configuration option in the settings and default to enabled if IPFS isn't currently found in the users path
- b) Incentivise them to enable it with a dialog (preferably after an acton is taken): If IPFS is already found, present a prompt explaining the benefit of using the desktop binary as default but with a sentence in the dialog saying: "This can be changed at any time from the settings page."
- c) Success 🥳

There may be other opportunities on the UI where we can present the option to enable this before an action is performed, such as first starting your node.

hacdias:
@cwaring I get you, but let me give you a little more context:

1. On Windows, we add it to the PATH anyway during installation. No action required from the user.
2. On Linux, we add it to the PATH **if** not found already during installation. No action required from the user. Can't be changed later via IPFS Desktop.
3. On macOS, we do the same as we do on Linux but **during runtime** because it's possible. This way we have more flexibility and can ask the user if they want to replace it.

I just wanted to get feedback from more people too. /cc @olizilla @fsdiogo

fsdiogo:
@hacdias my opinion hasn't changed => https://github.com/ipfs-shipyard/ipfs-desktop/pull/906#issuecomment-483704932.

hacdias:
So, just made some update: initially asks the user. If _no_, there will be an option under 'Advanced' to add to the PATH. It won't ask the user more times.

hacdias:
As an update: I'll use this PR and do what's described here (https://github.com/ipfs-shipyard/ipfs-desktop/issues/727#issuecomment-485438773) on this PR. So'll change the status to WIP

hacdias:
Ref for myself: https://github.com/jorangreef/sudo-prompt/issues/91

hacdias:
@olizilla @lidel could you please review the code and try it out? 😄

- On macOS:
 - No 'ipfs' + never asked --> add 'ipfs' to PATH
 - If 'ipfs' + never asked --> ask if they want to add
- On Linux --> ask if they want to add
- The option can be changed afterwards in Settings.

@lidel decided to use a fork of `sudo-prompt` due to this: https://github.com/jorangreef/sudo-prompt/issues/91.

This package will also unblock some issues on #911 if I implement it in a similar way. And we can use JS code ⚡️

src/late/ipfs-on-path/scripts/install.js
```diff
@@ -0,0 +1,46 @@
+const fs = require('fs-extra')
+const which = require('which')
+const { backup } = require('../../../utils/scripts/backup')
+const getArg = require('../../../utils/scripts/args')
+const { SOURCE_SCRIPT, DEST_SCRIPT } = require('./consts')
+
+// TODO: be careful of file permissions when fs.something()
```
hacdias: @lidel when using `sudo`, we need to be careful with the permissions when using the `fs.*` functions. Could you help me with that, please?

olizilla:
The current install prompts you with this

Screenshot 2019-04-23 at 09 45 31

As we only prompt the user about installing the `ipfs` command if they already have one, this message should talk about the need to replace your existing command... how about:

---

#### Allow IPFS Desktop to update the IPFS CLI

IPFS desktop can keep the IPFS command-line interface (CLI) up to date. It will replace the existing `ipfs` command installed at `/usr/local/bin/ipfs`, and update it when new versions are available.

[ Update CLI ] [ Keep existing ]

---

package.json
```diff
@@ -14,13 +14,9 @@
     "clean:webui": "shx rm -rf assets/webui/",
     "build": "run-s build:*",
     "build:webui": "run-s build:webui:*",
-    "build:webui:download": "npx ipfs-or-gateway -c QmfQkD8pBSBCBxWEwFSu4XaDVSWK6bjnNuaWZjMyQbyDub -p assets/webui/",
+    "build:webui:download": "npx ipfs-or-gateway -c QmQz6PVJPqRZZtShVZbc8J3VdZE8pEd7p5RGujvGGUmSyx -p assets/webui/",
```
olizilla: This CID should always be the latest release of ipfs-webui unless it depends on a pre-release version.

src/utils/scripts/backup.js
```diff
@@ -0,0 +1,59 @@
...
+      return
+    }
+
+    checkErr(err)
+  }
+
+  try {
+    fs.unlinkSync(backup)
+  } catch (_) {
+    // not a first world problem if we can't remove the backup
```
olizilla:
```suggestion
    // Failed to remove the backup. Suprising, but not worth bothering the user about.
```

src/utils/scripts/backup.js
```diff
@@ -0,0 +1,59 @@
...
+  const bin = original.split('/').pop()
+  const backup = join(userData, bin + '.bak')
+
+  try {
+    fs.copySync(original, backup)
+  } catch (err) {
+    checkErr(err)
+  }
+
+  console.log(`${original} copied to ${backup}`)
```
olizilla: Any reason not to use `logger` here?

src/utils/scripts/backup.js
```diff
@@ -0,0 +1,59 @@
...
+    fs.removeSync(dst)
+  } catch (err) {
+    if (err.code !== 'ENOENT') {
+      throw err
+    }
+  }
+
+  try {
+    fs.copySync(backup, dst)
+    console.log(`${backup} restored to ${dst}`)
```
olizilla: Any reason not to use `logger` here?

src/utils/scripts/backup.js
```diff
@@ -0,0 +1,59 @@
...
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      console.log(`${backup} not found, skipping`)
```
olizilla: Any reason not to use `logger` here?

src/late/ipfs-on-path/scripts/install.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+if (which.sync('node', { nothrow: true }) !== null) {
+  exists = true
+}
+
+if (exists) {
+  backup(getArg('user-data'), DEST_SCRIPT)
+  fs.unlinkSync(DEST_SCRIPT)
+  console.log(`${DEST_SCRIPT} removed`)
```
olizilla: Any reason not to use `logger` here?

assets/locales/en.json
```diff
@@ -42,5 +42,11 @@
   "runningLatestVersion": "You seem to be running the latest version.",
   "couldNotCheckForUpdates": "Could not check for updates",
   "pleaseCheckInternet": "Please check your Internet connection.",
-  "checkForUpdates": "Check for Updates..."
+  "checkForUpdates": "Check for Updates...",
+  "yes": "Yes",
+  "no": "No",
+  "anErrorHasOccurred": "An error has occurred",
+  "anUnexpectedErrorHasOccurred": "An unexpected error has occurred and it does not prevent the execution of IPFS Desktop. You can either inspect the logs for yourself or report the error to the developers.",
```
olizilla: We can simplify the language here but my intuition is "inspect the logs"  is almost never a useful action for anyone other than the developers of the app. 

```suggestion
  "anUnexpectedErrorHasOccurred": "An error occurred but it does not prevent the execution of IPFS Desktop. You can either inspect the logs for yourself or report the error to the developers.",
```

Perhaps this kind of error could be logged to analytics if enabled?

src/late/ipfs-on-path/index.js
```diff
@@ -0,0 +1,120 @@
...
+      '-win', 'hidden',
+      join(__dirname, `scripts/${script}.ps1`).replace('app.asar', 'app.asar.unpacked')
+    ], {}, err => {
+      if (err) {
+        logger.error(`[ipfs on path] ${err.toString()}`)
+        showRecoverableError(err)
+        return resolve(false)
+      }
+
+      logger.info(`[ipfs on path] ${script}ed`)
```
olizilla:
```suggestion
      logger.info(`[ipfs on path] ${script} installed`)
```

src/late/ipfs-on-path/scripts/install.js
```diff
@@ -0,0 +1,46 @@
...
+const { SOURCE_SCRIPT, DEST_SCRIPT } = require('./consts')
+
+// TODO: be careful of file permissions when fs.something()
+
+let exists = false
+
+try {
+  fs.lstatSync(DEST_SCRIPT)
+  exists = true
+  console.log(`${DEST_SCRIPT} already exists`)
```
olizilla: logger

olizilla: Oh i see, this script is excuted seperately...

src/late/ipfs-on-path/index.js
```diff
@@ -0,0 +1,120 @@
...
+      return false
+    }
+
+    logger.info(`[ipfs on path] ${stdout.toString().trim()}`)
+    return true
+  }
+
+  if (os.platform() === 'darwin') {
+    return new Promise(resolve => {
+      execFile(process.execPath, args, options, (err, stdout) => {
```
olizilla: So this runs IPFS Desktop as if it was `node` in order to run the install script?

package.json
hacdias: @olizilla it's from the linked PR (https://github.com/ipfs-shipyard/ipfs-webui/pull/1016) and needed, yes.

src/utils/scripts/backup.js
hacdias: As explained (not as well as it should) above:

> The **/scripts** directory are only to be used on this kind of external executed scripts. That's why I added them outside of ASAR.

So, imagine the *.js files in the `scripts` folder as... shell scripts. They're not linked to our Electron app since we're running Electron as Node. To use logger, we'd need access to Electron api.

src/utils/scripts/backup.js
hacdias: Same as above.

src/utils/scripts/backup.js
hacdias: Same as above.

src/late/ipfs-on-path/scripts/install.js
hacdias: Same as above.

src/late/ipfs-on-path/index.js
hacdias: `script` var is either `install` or `uninstall`. So we just add the `ed` in the end.

src/late/ipfs-on-path/index.js
hacdias: Exactly! Kinda, it runs Electron, it never runs IPFS Desktop code. Because Electron can run as Node. This way we don't have to care about shell scripts and just use node.

src/late/ipfs-on-path/scripts/install.js
hacdias: 😃 Yes, what I said above!

assets/locales/en.json
hacdias: Perhaps, we can think about that after this PR is finished.

What would you replace the inspect the logs part for?

hacdias:
> The current install prompts you with this
> 
> Screenshot 2019-04-23 at 09 45 31
> 
> As we only prompt the user about installing the `ipfs` command if they already have one, this message should talk about the need to replace your existing command... how about:
> 
> #### Allow IPFS Desktop to update the IPFS CLI
> IPFS desktop can keep the IPFS command-line interface (CLI) up to date. It will replace the existing `ipfs` command installed at `/usr/local/bin/ipfs`, and update it when new versions are available.
> 
> [ Update CLI ] [ Keep existing ]

Kinda. We also show it on Linux because there we always need permission. I'll update this later with a matrix of different messages.

hacdias:
@olizilla updated the messages, depending on whether or using Windows, or you already have `ipfs`, or you don't have it but we need permission.

src/late/ipfs-on-path/index.js
```diff
@@ -0,0 +1,125 @@
...
+        resolve(getResult(err, stdout))
+      })
+    })
+  }
+
+  options.name = 'IPFS Desktop'
+
+  const { err, stdout } = await sudo.exec(`${process.execPath} ${args.join(' ')}`, options)
```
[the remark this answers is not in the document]
hacdias: @lidel just added some more error handling: 44ffab94a0b5e726db9ed02675b7fe4c6e555737

Thanks for looking at it! 😃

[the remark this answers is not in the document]
lidel: @hacdias 44ffab9 still does not catch the polkit errors and results in hard crash:

> ![2019-04-25--14-51-30](https://user-images.githubusercontent.com/157609/56736905-a17c0f80-6769-11e9-9f18-5119fce3cc0e.png)

Try wrapping `await sudo.exec` with explicit  `try/catch`, as the promise can fail:

```diff
+   try {                                                                                                                                                                                                                                           
      const { err, stdout } = await sudo.exec(`${process.execPath} ${args.join(' ')}`, options)                                                                                                                                                     
      return getResult(err, stdout)                                                                                                                                                                                                                 
+   } catch (err) {                                                                                                                                                                                                                                 
+     return getResult(err)                                                                                                                                                                                                                         
+   }                                                                                                                                                                                                                                               
  }   
```

Also, messages about polkit are in `runWindows` for some reason, so I never see them :P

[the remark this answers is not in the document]
hacdias: Ah, oops! The code was so similar I put it in the wrong place 😅 Fixed: 22984ac60638ded14917d8b5b7e66a1eac66be42 😃

src/late/ipfs-on-path/scripts/install.js
hacdias: Should we `chown` to the user and group running IPFS Desktop process with which set of permissions?

lidel: `ipfs-on-path/scripts/install.js` will be executed as `root` via `await sudo.exec` in `ipfs-on-path/index.js`, which means it will have all permissions to create the symlink.

AFAIK a symlink itself does not have its own permissions (at least on Linux), instead it inherits permissions of original binary, so if user was able to execute `/some/path/to/ipfs` before, it will also work via `/usr/local/bin/ipfs`

In other words, I don't think we need to do anything additional here (but we need to test if symlinking actually works on a distribution with polkit)

hacdias:
I'm merging this and adding the [missing Web UI PR](https://github.com/ipfs-shipyard/ipfs-webui/pull/1016) to the list of the next release #880.

PR #909, opened 2019-04-19
feat: add snap hooks

hacdias:
So, this should add the snap hooks for install and removal to add `ipfs` to the `PATH`.

Although, building is failing with "Macro location is not defined". I couldn't find any information on Electron Builder's issues nor on Google. Everything I find is related to C... so... I'm a bit confused with what's happening.

@lidel could you take a look at the code to see if something looks wrong? I'll keep searching to try to find out what's wrong. It seems related to fpm, but I don't change any of its configuration here...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 config.gypi
 assets/webui
 *.nupkg
+assets/build/snap-hooks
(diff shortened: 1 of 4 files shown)


hacdias:
Uuhhh, very much thanks @lidel 😃

Issue #907, opened 2019-04-16
Add opt-in npm-on-ipfs support


olizilla:
As per discussions over on https://github.com/ipfs-shipyard/npm-on-ipfs/issues/94 we'd like make it easy for ipfs-desktop users to try out [npm-on-ipfs](https://github.com/ipfs-shipyard/npm-on-ipfs), and co-host modules that they install.

- [ ] Offer a prompt to install the `ipfs-npm` command line tools
- [ ] alias it as `npm` or `yarn`

olizilla:
There are some UX challenges to solve here.

- Should be easy to opt in to using `ipfs-npm` in place of `npm`. Perhaps checkbox in the settings page is the simplest initial implementation
- Should be easy to undo. If you want to reset back to regular  `npm` then you should be able to just uncheck the box.

I'm sure there are more. @hacdias @ericronne drop any thoughts you have here!

hacdias:
I agree with the checkbox in settings. Also, I think that on a first run, if we detect that Node.js is installed, we could actively prompt the user.

hacdias:
So I was implementing this and an issue surged: I was thinking about aliasing `ipfs-npm` as `npm` and to do so I thought about replacing /usr/local/bin/npm with our script. Although this would create a loop on ipfs-npm where it'd call itself forever when looking up for npm.

We could set an `alias` on the user profile but this would require (1) finding the shell they're using and (2) finding a way to do this on Windows.

The other option I see is trying to set `--package-manager` to the right path.

lidel:
> I was thinking about aliasing ipfs-npm as npm and to do so I thought about replacing /usr/local/bin/npm with our script

This feels risky. Quick sanity check: is **moving binaries belonging to other packages installed in the system** is better than **adding a directory in front of `PATH` variable**?
Think about scenario when someone uninstalls `npm`. Our proxy be removed. 

Unless I am, missing something, 
adding a directory to the front of `PATH` seems the only "safe" way to override.

hacdias:
@lidel I agree that it is risky changing binaries. For now, we decided just to install `npm-on-ipfs` package and tell the user to use `ipfs-npm` instead of regular `npm/yarn` commands. This needs some more thought because changing the variable PATH is not as easy as it seems since it might depend a lot on the shell the user is using.

On Windows it's relatively easy and we already do it to add IPFS to the PATH.

DavidBurela:
[...]

PR #911, opened 2019-04-21
feat: npm on ipfs

hacdias:
Please see https://github.com/ipfs-shipyard/ipfs-desktop/issues/907#issuecomment-485283109 (went with the last option).

- [x] Merge https://github.com/ipfs-shipyard/ipfs-webui/pull/1048
- [x] Update Web UI hash


hacdias:
@lidel @olizilla could you go over the code, see my `TODO`s and tell me your opinion about those topics, please? Also, if you find something else that looks odd, ask. I want this working the best it can and as reversible as possible. 😃

src/late/npm-on-ipfs/index.js
```diff
@@ -0,0 +1,54 @@
...
+export default function (ctx) {
+  run()
+
+  const activate = (value, oldValue) => {
+    if (value === oldValue) return
+    if (value === true) return install()
+    return uninstall()
+  }
+
+  createToggler(ctx, 'npmOnIpfs', activate)
```
hacdias: NOTE: This has a bug. Any change will trigger this even if do it bellow. **CHECK**

hacdias:
One more note: since we're adding `sudo-prompt` to #906 (see comments), then we can overcome the permissions issue on Linux here with it too.

olizilla:
@hacdias did you want to update this PR now that #906 has been merged?

hacdias:
@olizilla I do and I will 😃

src/late/npm-on-ipfs/scripts/npm.js
```diff
@@ -0,0 +1,24 @@
...
+const { readFile } = require('fs')
+const { execFile } = require('child_process')
+const { join } = require('path')
+
+readFile(join(homedir, '.ipfs/api'), (err, data) => {
+  const pkg = process.argv[1].includes('npm') ? 'npm' : 'yarn'
+  const addr = data.toString()
+  let args = [
+    ...process.argv.slice(2, process.argv.length),
+    '--package-manager', join(__dirname, `./${pkg}.bak`)
```
hacdias: @olizilla @lidel I need your advise here. We need to tell our `npm-on-ipfs` binary where is the actual package manager to use. The issue is: where should we back it up?

For the IPFS CLI tools, we back them up to IPFS Desktop user data directory, which won't be removed during any installations/uninstallation, which is good. And we can pass that directory path along to the scripts.

Although, in this case, `npm.js` will be executed plainly as `npm` or `yarn` so we can't tell where the user directory is. One possibility I see is to add to `/usr/local/bin` a file with the hardcoded path to the yarn and npm backups and we could keep this file as template.

Right now I'm just throwing it in the source directory, which is neither a good practice, nor useful since it would be removed by deleting the app. What do you think? **Where should we back npm/yarn to?** And how to handle that?

lidel: Is the underlying problem "I don't know the path to `npm`" ?

On linux/mac there are two handy commands:
- `where npm` shows _all_ `npm` binaries across directories listed in `PATH` 
- `which npm` shows which binary is selected when you execute `npm`

I believe `where` is also on Windows.

(If I misunderstood your underlying problem, I'll need an example to illustrate it)

olizilla: I think the question is: 

_if we move the original `npm` or  `yarn`  commands off the `PATH`, to replace them with the `ipfs-npm` proxy, where should we stash them so we can revert the change later?_

hacdias: The question is a mix of what @olizilla said and something else: where should we stash them so we can tell ipfs-npm to use them? Because we need to pass the original binary through `package-manager`, otherwise it will end up in a loop.

olizilla: gotcha... I'd do something like move it to `_npm`. it feels gross though. Worth some further thinking... like what are the other options

1. dont mess with the users PATH... add `ipfs-npm` but swap out `npm`
2. move `npm` to `_npm` so we can still call  it from `ipfs-npm`
3. bundle a version of `npm`  with `ipfs-npm` so it can reliably call that one.

The more I think about the worse option 2 sounds. I would go with option 1 for now, which simplifies this PR and makes it less risky. Also, we still need to do work on `ipfs-npm` to  make it more user friendly, so it make sense to **not** swap it out for the regular `npm` yet... it's still experimental, so it's not ready for prime time yet.

Once it is more stable, option 3 might be a better way to go, so that ipfs-npm is proxying a known version of `npm` rather than whatever the user happens to have installed.

hacdias: So you're suggesting just to install npm-ipfs package right?

lidel: My 2 cents: remember to do a quick sanity check when deciding how to override `npm.

Namely, is **moving binaries belonging to other packages installed in the system** is better than **adding a directory in front of `PATH` variable**?
Think about scenario when someone uninstalls `npm`. Our proxy be removed. 

Unless I am, missing something, 
adding a directory to the front of `PATH` seems the only "safe" way to override.

hacdias:
So we decided to just install `ipfs-npm` package with this PR and _not_ change anything else in the system. For what is worth, the PR looks finished. We just need a toggler in the UI to install/uninstall `ipfs-npm`.

@lidel I think we already talked about this: I'm running `npm (un)install -g` as the user that's running IPFS Desktop. That might not work on all environments. It should work on Windows and macOS, but what about Linux?

lidel:
@hacdias I don't want us to spend too much time debugging fringe distributions. 
It should work for most people using mainstream defaults, and power users will not mind doing it manually.  Just make sure there is proper error handling when `npm install -g ipfs-npm` fails, asking user to run "npm i -g ipfs-npm" manually.  :ok_hand:

hacdias:
Completely agree @lidel! Thanks.

So, for this PR to be merged, we only need to wait for https://github.com/ipfs-shipyard/ipfs-webui/pull/1048 and decide how to send the errors and change the options. Otherwise, it already works.

src/late/npm-on-ipfs/index.js
```diff
@@ -0,0 +1,42 @@
...
+    interval = setInterval(pkg.update, 43200000)
+  }
+
+  if (opt !== null) {
+    logger.info('[npm on ipfs] no action taken')
+    return
+  }
+
+  // First time running this function.
+  if (!which.sync('node', { nothrow: true })) {
```
olizilla: minor, but we should check if `npm` is available  rather than `node`. We are relying on `npm` existing later on and it's getting increasingly possible that someone would have node, but use a different pacakge manager.

src/late/npm-on-ipfs/package.js
```diff
@@ -0,0 +1,46 @@
...
+    // NOTE: might fail on older NPM (< 6.9.1) versions
+    // https://github.com/npm/cli/pull/173
+    const { stdout } = await execFile('npm', ['outdated', '-g'])
+
+    if (stdout.indexOf('ipfs-npm') === -1) {
+      logger.info('[npm on ipfs] ipfs-npm: is up to date')
+      return
+    }
+  } catch (e) {
+    logger.error('[npm on ipfs] ipfs-npm: could not check if up to date %v', e)
```
olizilla: Maybe throw? we shouldn't let it fall through the next log line.

hacdias: My idea here was to ignore this error and update it anyways. It will fail for most people due to a bug on npm, which was not yet released: https://github.com/ipfs-shipyard/ipfs-desktop/blob/feat/npm-on-ipfs/src/late/npm-on-ipfs/package.js#L10-L11

hacdias:
@olizilla could you take one last look at this please?

hacdias:
Two issues arising on Travis ATM not related to the PR:

- https://benlimmer.com/2019/01/14/travis-ci-xvfb/
- `dist.ipfs.io` timing out.

src/late/npm-on-ipfs/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+  // Confirms if the package is still (un)installed because the user
+  // might change it manually.
+  if (opt !== null || exists !== false) {
+    store.set(SETTINGS_OPTION, exists)
+    opt = exists
+  }
+
+  if (opt === true) {
+    interval = setInterval(existsAndUpdate, 43200000)
```
fsdiogo:
```suggestion
    interval = setInterval(existsAndUpdate, 43200000) // every 12 hours
```

hacdias:
Travis also failing on master with timeout getting go-ipfs binary. I'm merging this.

andrew:
:tada:

PR #927, opened 2019-05-13
fix: add folder

### LINKED ISSUE #884: Using "add folder" does not add a folder correctly [0.7.2]
- **OS**: win32
- **IPFS Desktop Version**: 0.7.2
- **Electron Version**: 4.0.8
- **Chrome Version**: 69.0.3497.128

Dragging a folder into the web ui, will add it correctly.
But clicking "add folder" seems to add an empty folder, that if you try to navigate into will result in it just continuously trying to load.

I have included a screencapture gif to show the bug.

BF29E87D-D9A4-4005-914E-DF3BEEA09E28

hacdias:
Fixes #884.

Requires https://github.com/ipfs-shipyard/ipfs-webui/pull/1034/files.


hacdias:
When testing, don't forget to run `npm run clean:webui` and `npm run build:webui` to install the new version to test.

src/late/webui/preload.js
```diff
@@ -31,6 +35,37 @@ window.ipfsDesktop = {
...
+    return new Promise(resolve => {
+      remote.dialog.showOpenDialog(remote.getCurrentWindow(), {
+        title: 'Select a directory',
+        properties: [
+          'openDirectory',
+          'createDirectory'
+        ]
+      }, async (res) => {
+        if (!res || res.length === 0) {
+          return resolve()
```
olizilla: Would it be useful to make that a reject, so calling code can tell when you've cancelled? Testing out this PR the current behaviour if you hit cancel from the file dialog it leaves the drop down menu open, but if you cancel from the directory dialog it closes the drop down. I think it should close the drop down in both cases.

hacdias: The issue's on Web UI code. I'll apply a fix on Web UI soon. We shouldn't reject I think because on the browser side we just receive a `null` variable, not a rejection.

hacdias: See https://github.com/ipfs-shipyard/ipfs-webui/pull/1051

PR #932, opened 2019-05-21
fix: ensure strict cors checking is enabled on api

olizilla:
The default config created by ipfsd-ctl allows cross-origin requests from any domain. This PR ensures that ipfs-desktop creates a new ipfs config with strict CORS checking for API requests.

There is also a one-time check to fix existing config files where we find ones with CORS checking disabled.

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>


olizilla:
Related: https://github.com/ipfs/js-ipfsd-ctl/issues/333

olizilla:
@lidel it worked. Can I get your thoughts on the PR now CI is happy.

olizilla:
@lidel that was deliberate, but I agree, I think webui should fix it, at least on first run. 

As a seperate piece of work the webui should warn folks when it finds  the wildcard in the allowed origins config. That way if it gets added back later, you still get informed about it.

olizilla:
@lidel @hacdias can you take another look please. I've chanaged it so it'll remove `*` regardless of what other domains are in there. It only does it once. If you add it back later it will leave it.

I've also ensured that it doesn't blow up if the config is invalid... that is something we should handle  gracefully elsehwhere. I'll raise an issue.

src/utils/daemon.js
```diff
@@ -106,20 +106,32 @@ function writeConfigFile (ipfsd, config) {
...
+// Check for * and webui://- in allowed origins on API headers.
+// THe wildcard was a ipfsd-ctl default, that we dont want, and webui://- was an
+// earlier experiement that should be cleared out.
+//
+// We remove them the first time we find them. If we find it again on subsequent
+// runs then we leave them in, under the assumption that you really want it.
+// TODO: show warning in UI when wildcard is in the allowed origins.
 function checkCorsConfig (ipfsd) {
-  const updatedCorsConfig = store.get('updatedCorsConfig')
+  if (store.get('checkedCorsConfig')) {
```
hacdias: We shouldn't rely on stored values. What if you set up an IPFS node yourself and configure it yourself and then you start using IPFS Desktop with that repository? Will we change the configuration without the user knowing? Perhaps we should only do this on `ipfsd.init`.

olizilla: I am ok with that risk. I want to fix older config files for users who have installed ipfs-desktop already. The previous iteration of this PR was much more likely to only run for users who had installed desktop and hadn'd edited their API config block. But based on @lidel's feedback, it's now more proactive and will remove the wildcard even if you have added additional domains.

lidel: Agreed. The end goal is to do this only on `init` of new repo, but for now we need to proactively fix security issue opened by default config in  `ipfsd`. 

It should  stay that way for a release or two to ensure all users  got the fix. Then we can make it more passive.

hacdias:
Per both reviews, I'm merging this @olizilla. Thanks for fixing it 😃

olizilla:
@hacdias i'm not finished here. I'm in the middle of squashing and backporting this change to a branch off of the 0.7.2 tag. I guess i can push to a new branch for that purpose.

olizilla:
ah, is ok, is squased on merge, so i can just cherry pick from there. huzzah!

PR #942, opened 2019-05-30
feat: restart daemon when connection changes

### LINKED ISSUE #930: Restart ipfs daemon when network address changes
If you start IPFS on network A, and then move to network B, there is nothing in libp2p or go-ipfs to automatically detect that and update the swarm addresses you are announcing.
See: 
- https://github.com/ipfs/community/issues/408#issuecomment-490032167

IPFS Desktop could detect a network change and restart the deamon automatically.

hacdias:
Restarts the daemon when the connection gets changed. To test it, try turning off and on the WiFi or switching network.

Closes #930.

License: MIT


olizilla:
This is neat. I think the general pattern should be to inform the main process what happened... so it should send an event like `online-status-changed` (taken from the [docs](https://electronjs.org/docs/tutorial/online-offline-events)) and the main process should decide what to do about it, in this case, restart. Right now the UI is telling the main process what to do, not what happened.

hacdias:
@olizilla just made the changes you've asked and working 😃 We're comparing to `false` to make sure it isn't triggered the first time we get the status. I'm going to merge this.

PR #943, opened 2019-05-31
feat: add restart ipfs function

### LINKED ISSUE #920: Save config and reload daemon automatically
We can add a function to IPFS Desktop bundle to tell IPFS Desktop to restart the daemon when changes are made to the settings.

hacdias:
Adds `restartIpfs` function to `window.ipfsDesktop` object.

- See https://github.com/ipfs-shipyard/ipfs-webui/pull/1050

License: MIT


olizilla:
Rather than letting the UI trigger restarts directly, the main process should probably listen for config change events and decide if it should restart... it might choose not to becuase a restart is in progress or a (fictional) user setting disables it.

hacdias:
@olizilla just made the changes. It now listens for config changes and then restarts.

PR #963, opened 2019-06-21
chore: update dependencies

hacdias:


hacdias:
Merging. Travis failing because it can't download `go-ipfs`.

Issue #974, opened 2019-07-14
IPFS not getting (always) killed when sutting down


hacdias:
I'm having lots of issues of IPFS not getting actually killed when shutting down IPFS Desktop. IPFS Companion shows it as running with 0 peers and I need to kill it via `killall ipfs`.

PR #978, opened 2019-07-15
fix: stopping daemon with new ipfsd

hacdias:
Fixes the stop daemon function. Forgot to change to the async API here.

License: MIT


hacdias:
Wonder if this was causing #974. Will wait to see.

PR #989, opened 2019-07-17
fix: handle port collision gracefully

### LINKED ISSUE #912: IPFS has errored, because 8080 is already in use
I just downloaded the AppImage from the Readme, and it won't connect to IPFS because, as described in the log, my port 8080 is already in use.

```
2019-04-22T14:52:31.218Z error: [ipfsd] %vCommand failed: /tmp/.mount_ipfs-d8VxVBQ/resources/app.asar.unpacked/node_modules/go-ipfs-dep/go-ipfs/ipfs daemon --migrate=true --routing=dhtclient --enable-gc=true
Error: serveHTTPGateway: manet.Listen(/ip4/127.0.0.1/tcp/8080) failed: listen tcp4 127.0.0.1:8080: bind: address already in use
```

This port is very common and is even used by Skype, so maybe you could use another port if 8080 is taken? 

Thank you

hacdias:
Fixes #912.

License: MIT


hacdias:
@lidel @olizilla  could you take a look at what I've done so far, please? What do you think we should alert the user for this?

hacdias:
And should we revert the ports when shutting down the daemon? What if it is not possible?

src/utils/daemon.js
```diff
@@ -139,10 +142,82 @@ function checkCorsConfig (ipfsd) {
...
+
+  const isApiMaDaemon = await checkIfAddrIsDaemon(configApiMa.nodeAddress())
+  const isGatewayMaDaemon = await checkIfAddrIsDaemon(configGatewayMa.nodeAddress())
+
+  if (isApiMaDaemon && isGatewayMaDaemon) {
+    // TODO: BEING USED BY A DAEMON
+    return
+  }
+
+  const apiPort = parseInt(configApiMa.nodeAddress().port)
```
lidel: JS uses [fancy heuristic to calculate radix based on input](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description), it is usually safer to provide explicit one:
```suggestion
  const apiPort = parseInt(configApiMa.nodeAddress().port, 10)
```

src/utils/daemon.js
```diff
@@ -139,10 +142,82 @@ function checkCorsConfig (ipfsd) {
...
+  const gatewayPort = parseInt(configGatewayMa.nodeAddress().port)
```
lidel:
```suggestion
  const gatewayPort = parseInt(configGatewayMa.nodeAddress().port, 10)
```

src/utils/daemon.js
```diff
@@ -139,10 +142,82 @@ function checkCorsConfig (ipfsd) {
...
+    config.Addresses.API = config.Addresses.API.replace(apiPort.toString(), freeApiPort.toString())
+  }
+
+  if (gatewayPort !== freeGatewayPort) {
+    config.Addresses.Gateway = config.Addresses.Gateway.replace(gatewayPort.toString(), freeGatewayPort.toString())
+  }
+
+  // TODO: alert user? revert when shutting down? How?
+
+  writeConfigFile(ipfsd, config)
```
lidel: Probably should not touch config file if port did not change.

src/utils/daemon.js
```diff
@@ -139,10 +142,82 @@ function checkCorsConfig (ipfsd) {
...
+
+  if (apiPort !== freeApiPort) {
+    config.Addresses.API = config.Addresses.API.replace(apiPort.toString(), freeApiPort.toString())
+  }
+
+  if (gatewayPort !== freeGatewayPort) {
+    config.Addresses.Gateway = config.Addresses.Gateway.replace(gatewayPort.toString(), freeGatewayPort.toString())
+  }
+
+  // TODO: alert user? revert when shutting down? How?
```
lidel: I think ipfs-desktop should inform user the port X is not available and show buttons with options to continue:
- **Use port Y instead** (updates config and tries to start again)
- **Try again** (does not change config, repeats the `getPort` test – useful for user who wants to shut down the thing that took the port without breaking the flow)
- **Quit** (does not change config, exits)

Once we change the config, we don't revert it.

hacdias:
So I think this is ready for review: everything is logged and we alert the user when one or two ports are busy and give them a choice.

image


image

If they choose to close the dialog, IPFS just doesn't try to turn on the daemon and the menubar stays like this:

image

assets/locales/en.json
```diff
@@ -77,5 +77,15 @@
   "ipfsDesktopHasShutdownDialog": {
     "title": "IPFS Desktop has shutdown",
     "message": "IPFS Desktop has shutdown because of an error. You can restart the app or report the error to the developers, which requires a GitHub account."
+  },
+  "busyPortsDialog": {
+    "title": "Ports are busy",
+    "message": "The ports { port1 } and { port2 } are not available. Do you want to use ports { alt1 } and { alt2 } instead?",
```
lidel:
```suggestion
    "message": "The ports { port1 } and { port2 } are not available. Do you want to use { alt1 } and { alt2 } instead?",
```

PR #990, opened 2019-07-17
feat: allow to move repository

### LINKED ISSUE #988: Prompt to set IPFS_PATH During Installation
The .ipfs folder becomes very large and is better suited to a storage drive than the operating system's drive, where possible.  A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

Prompt newbies to pick their storage drive for this purpose during installation and open a GUI so they can navigate to a suitable drive's folder.

**Describe alternatives you've considered**
Manually editing the config is complex. The .ipfs folder's location is not listed in the config, so it can't easily be found or changed.

**Additional context**
Is this compatible with Open Bazzar? Will installing them both result in two large ipfs folders?

hacdias:
Closes #988.

License: MIT


hacdias:
This adds an option under the Advanced menu that allows the user to change the directory of their repository.

hacdias:
image

image

image

hacdias:
> I would rename "OK" on the first popup to "Move and restart node", to communicate what will happen when clicked.

Done!

> I was bit surprised the feature is not listed on Web UI Settings screen, but Advanced submenu (makes it harder to discover, but maybe it is a good thing?)

All the 'Advanced' options we have are not shown within the Web UI Settings configuration. It's just a matter of deciding where this should be. Although, I'd think this is an advanced config.

> Some options:

I vote on symlinking 😃 I just need to make sure if we need (or not) elevated permissions on Windows

olizilla:
I think desktop should set the `IPFS_PATH` variable if the user elects to have ipfs-desktop provide `ipfs` on the `PATH`, and as such should keep it up to date if the user moves their repo.

hacdias:
@olizilla on Windows that'd be quite easy to do, but on macOS/Linux I can't see an easy way to do so because it all depends on the shell you're using and possibly profiles. What do you think? /cc @lidel

hacdias:
Ping @olizilla @lidel

lidel:
Indeed, setting `IPFS_PATH` on Linux is tricky (different users, shells, conventions).

@hacdias  I believe the cleanest way to add an env variable globally on Linux is to drop `ipfs.sh` into `/etc/profile.d/` directory. It needs to be executable.
In most of the modern distributions `/etc/profile` will init variables by executing scripts from those files. 
It makes it cleaner than adding lines to `/etc/profile` or other place directly.

@olizilla would the same work on macOS? can you check if `/etc/profile` has something like

```bash
# process *.sh files in /etc/profile.d
for sh in /etc/profile.d/*.sh ; do
	[ -r "$sh" ] && . "$sh"
done
unset sh
```

hacdias:
@lidel I'm also using macOS for a few months already 😃 Here's the content of `/etc/profile`:

```
# System-wide .profile for sh(1)

if [ -x /usr/libexec/path_helper ]; then
	eval `/usr/libexec/path_helper -s`
fi

if [ "${BASH-no}" != "no" ]; then
	[ -r /etc/bashrc ] && . /etc/bashrc
fi
```

So it doesn't seem to load anything from that folder

lidel:
Would be cool to limit the complexity and use `/etc/profile.d` on all unix-like systems.

On Mac you could add code that adds support for `/etc/profile.d/*.sh` to `/etc/profile` if missing.

A simple check could be `grep /etc/profile.d /etc/profile` (just to see if `profile.d` is referenced in the file)

hacdias:
@lidel If we did that, do you think we should still keep adding the binary to `/usr/local/bin` or we should just append our bin path to the PATH via a file in `/etc/profile.d`?

lidel:
@hacdias Let's keep the scope of this PR limited to `IPFS_PATH`.

If we make it work on both mac and linux, we can evaluate adding `PATH` and removal of the the symlink.
There may be unknown unknowns on mac, so it should be a separate issue/PR.

olizilla:
for macos, we should investigate the .plist file https://apple.stackexchange.com/questions/267025/set-environment-variable-for-applications-on-startup

olizilla:
given that when ipfs-desktop provides `ipfs` it is actually pointing to a shell script rather than the ipfs binary directly, could we `export IPFS_PATH=<new path>` from inside that script?

hacdias:
That would, perhaps, be the easiest and simplest way to do it. What do you think @lidel?

lidel:
Agree, adding that one line will be easier way to get this PR merged.

If we want to remove proxy script and stop symlinking, we can investigate `/etc/profile.d` in separate PR.

hacdias:
Updates done! Works on Windows and macOS. @lidel could you take a look to see if it sounds/looks good?

src/ipfs-on-path/scripts/ipfs.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+export IPFS_PATH="/Users/henriquedias/Desktop/.ipfs"
```
lidel: We can ask all users to rename to `henriquedias` :upside_down_face:  but maybe read the value from ipfs-desktop's `config.json`/`ipfsConfig.path` instead ;)

hacdias: 😆 This was... unfortunate and funny. The thing is: from where should we read this? How to do it on windows? How to parse a json file? Where is the json file? There's lot of variables we don't know and we'd always need an hardcoded thing. Perhaps we could have a file within that directory that is gitignored and contains the path. What do you think?

hacdias:
Tested on macOS, Windows and Ubuntu. Worked. Merging.

woss:
i tried on windows 10 and it works only if the new repo location is inside the folder. the root of the HDD doesn't work. so `D:/ipfs` works but `D:/` doesn't

PR #1038, opened 2019-08-16
feat: write some unit tests

hacdias:


hacdias:
@olizilla could you give me some guidance here? If this tests look good, I'd like to merge them. I wonder if I'm playing correctly with the mocks: if I keep this way, I'll need to do lots and lots of mocks for lots of things we import. What do you think?

hacdias:
Ping @olizilla

PR #1068, opened 2019-08-30
feat: analytics

### LINKED ISSUE #1063: Analytics
We should add new analytics information. We could use the categories we already have on IPFS Web UI such as sessions and events and use the same information. Data to collect:

## Sessions

- Start
- Stop

## Events

- Daemon started
- Daemon stopped
- Web UI open
- Web UI closed
- Screenshot taken
- Download hash

## App errors

- Well... errors.

---

This will hopefully give us more insight about how the application is used and which features are more used.

hacdias:
- Closes #1063.
- Requires merging https://github.com/ipfs-shipyard/ipfs-webui/pull/1136.

This adds Analytics to IPFS Desktop using our Countly setup. This PR, in pair with the one on [IPFS Web UI repository](https://github.com/ipfs-shipyard/ipfs-webui/pull/1136), make a seamless integration between both Countly setups: we share the device id and the consents so we know on both sides what we can and can't send.

For this to work, I just created some wrappers around the logging functions to allow to send an event with it. If it has an event (`withAnalytics`), we log the event to Countly.

Also, created a `logger.start` function that allows us to measure, log and track time of certain actions, such as starting and stopping the daemon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2541,6 +2541,15 @@
       "integrity": "sha1-BuuE8A7qQT2oav/vrL/7Ngk7PFA=",
       "dev": true
     },
+    "camel-case": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/camel-case/-/camel-case-3.0.0.tgz",
+      "integrity": "sha1-yjw2iKTpzzpM2nd9xNy8cTJJz3M=",
+      "requires": {
+        "no-case": "^2.2.0",
+        "upper-case": "^1.1.1"
+      }
+    },
     "camelcase": {
       "version": "5.3.1",
       "resolved": "https://registry.npmjs.org/camelcase/-/camelcase-5.3.1.tgz",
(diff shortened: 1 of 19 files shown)


hacdias:
@olizilla the requested changes were made:

1. We now track sessions on Web UI as we did.
2. We now provide the list of the events we track to Web UI so they can be shown to the user.
3. And I also moved the feature enabling/disabling tracking to Web UI (when the user clicks to toggle options).

hacdias:
@olizilla after we talked, I tried moving all the logic to Web UI side to keep the countly id and such the same so there wouldn't be any problems. Although, that way, since the Web UI sometimes reloads (when the daemon stops and starts) or due to other reasons (user forced reload, we updated the files list), we won't be able to capture all events. Some of them are quite important such as daemon's starting time.

Although, I found out we could actually merge the [old ID data with the new](https://resources.count.ly/docs/countly-sdk-for-web#section-changing-device-id): https://github.com/ipfs-shipyard/ipfs-webui/pull/1136/commits/d20980f495e64ac187f5ca5bb629c3fd86876606.

Waiting for a review now :)

PR #1071, opened 2019-09-02
fix: increased window draggable area

### LINKED ISSUE #1066: More precise affordance to drag IPFS webUI window around
image

When I want to move around the webUI window, I usually try to click and hold anywhere in the light teal area at the top of the window to move it - however only the top 1/3 of that teal area is actually able to drag the window, moving my cursor while clicking anywhere else will highlight text in the window. Similarly, if I click in the upper 1/2 of the white text entry box to explore a hash, I don't focus the text entry area but start dragging the window. Can we do a better job highlighting the affordance to drag this window effectively?

I'm on MacOS Mojave 10.14.4

hacdias:
Increases the draggable area of IPFS Desktop. The new draggable area is (anywhere in the light blue area can be clicked to drag and move the window around):

image

It also fixes the issues when focusing the buttons/input.

- Needs https://github.com/ipfs-shipyard/ipfs-webui/pull/1138
- Fixes #1066

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "clean:webui": "shx rm -rf assets/webui/",
     "build": "run-s build:*",
     "build:webui": "run-s build:webui:*",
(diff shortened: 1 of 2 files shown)


hacdias:
@olizilla I decided to merge this since we already talked about it on a call and it's quite trivial. Anything you think that might be worth changing, let me know.

PR #1092, opened 2019-09-06
fix: allow multiple gateway and api ports

### LINKED ISSUE #1091: Error when have multiple API or Gateway addresses
```
2019-05-24T12:07:10.926Z error: Error: 36:68: execution error: System Events got an error: Can’t get login item "IPFS Desktop". (-1728)

    at ChildProcess.<anonymous> (/Applications/IPFS Desktop.app/Contents/Resources/app.asar/node_modules/applescript/lib/applescript.js:49:13)
    at ChildProcess.emit (events.js:182:13)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:12)
2019-09-06T11:36:35.896Z error: [ipfsd] addr must be a string, Buffer, or another Multiaddr
```

hacdias:
Fixes #1091

On the case there is an array of API or Gateway addresses and one or more of them collide with already busy ports, we show this message:

image

Users that have multiple addresses for API/Gateway already changed the configuration on their own and know how to solve this kind of issues.

License: MIT

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -109,6 +109,6 @@
   "multipleBusyPortsDialog": {
     "title": "Ports are busy",
     "message": "Cannot bind to one or more of the API or Gateway addresses because the ports are busy.",
-    "action": "I will fix the configuration"
+    "action": "Open configuration file"
   }
 }
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -105,5 +105,10 @@
     "title": "Port is busy",
     "message": "The port { port } is not available. Do you want to use { alt } instead?",
     "action": "Use port { alt } instead"
+  },
+  "multipleBusyPortsDialog": {
+    "title": "Ports are busy",
+    "message": "Some of the API or Gateway addresses cannot be binded to because ports are busy.",
+    "action": "I will fix the configuration"
   }
 }
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -108,7 +108,7 @@
   },
   "multipleBusyPortsDialog": {
     "title": "Ports are busy",
-    "message": "Some of the API or Gateway addresses cannot be binded to because ports are busy.",
+    "message": "Cannot bind to one or more of the API or Gateway addresses because the ports are busy.",
     "action": "I will fix the configuration"
   }
 }
--- a/src/daemon/daemon.js
+++ b/src/daemon/daemon.js
@@ -169,9 +169,16 @@ const parseCfgMultiaddr = (addr) => (addr.includes('/http')
 )
 
 async function checkPortsArray (addrs) {
+  addrs = addrs.filter(Boolean)
+
   for (const addr of addrs) {
     const ma = parseCfgMultiaddr(addr)
     const port = parseInt(ma.nodeAddress().port, 10)
+
+    if (port === 0) {
+      continue
+    }
+
     const isDaemon = await checkIfAddrIsDaemon(ma.nodeAddress())
 
     if (isDaemon) {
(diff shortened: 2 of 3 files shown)


src/daemon/daemon.js
```diff
@@ -168,9 +168,53 @@ const parseCfgMultiaddr = (addr) => (addr.includes('/http')
...
+
+    let addrs = apiIsArr
+      ? config.Addresses.API
+      : [config.Addresses.API]
+
+    addrs = gatewayIsArr
+      ? addrs.concat(config.Addresses.Gateway)
+      : addrs.concat([config.Addresses.Gateway])
+
+    return checkPortsArray(addrs)
```
alanshaw: Up to you if you want to change but you can simplify this loads to just:

```suggestion
    return checkPortsArray([].concat(config.Addresses.API, config.Addresses.Gateway))
```

You can pass arrays to `concat` and it'll concat all the items so it doesn't matter if the config is an array or not:

```js
> const foo = [1, 2, 3]
undefined
> foo.concat([4, 5, 6], 7, [8, 9, 10])
[
  1, 2, 3, 4, 5,
  6, 7, 8, 9, 10
]
```

src/daemon/daemon.js
```diff
@@ -168,9 +168,53 @@ const parseCfgMultiaddr = (addr) => (addr.includes('/http')
   : multiaddr(addr).encapsulate('/http')
 )
 
+async function checkPortsArray (addrs) {
+  for (const addr of addrs) {
+    const ma = parseCfgMultiaddr(addr)
+    const port = parseInt(ma.nodeAddress().port, 10)
```
alanshaw: You probably need to check if the port is 0 here and skip the free port check.

src/daemon/daemon.js
```diff
@@ -168,9 +168,53 @@ const parseCfgMultiaddr = (addr) => (addr.includes('/http')
   : multiaddr(addr).encapsulate('/http')
 )
 
+async function checkPortsArray (addrs) {
+  for (const addr of addrs) {
```
alanshaw: Filter `null`/`''`? I think go-ipfs errors on startup if you do this but in js-ipfs you can completely disable the Gateway/API from starting in this way.

```suggestion
  addrs = addrs.filter(Boolean)
  for (const addr of addrs) {
```

assets/locales/en.json
```diff
@@ -105,5 +105,10 @@
     "title": "Port is busy",
     "message": "The port { port } is not available. Do you want to use { alt } instead?",
     "action": "Use port { alt } instead"
+  },
+  "multipleBusyPortsDialog": {
+    "title": "Ports are busy",
+    "message": "Some of the API or Gateway addresses cannot be binded to because ports are busy.",
```
alanshaw: `bound` or `bind`, I think `binded` is not a word!

```suggestion
    "message": "Cannot bind to one or more of the API or Gateway addresses because the ports are busy.",
```

assets/locales/en.json
```diff
@@ -105,5 +105,10 @@
...
+    "action": "I will fix the configuration"
```
alanshaw: Suggestion: Open the config file in my editor?

assets/locales/en.json
hacdias: Ahah, thanks!!

assets/locales/en.json
hacdias: Close & Open Config File sounds good?

hacdias:
> Needs a test?

Just added one to make sure it boots up with multiple addresses. It does not test the case when there are collisions. Will do that in the feature.

There are a lot of tests planned!

hacdias:
As per @alanshaw's suggestion, the dialog was changed:

image

PR #1157, opened 2019-09-30
chore: snap stable channel

hacdias:
License: MIT


lidel:
Note:  this stopped attaching `.snap` to github release and instead uploads it to the snap store instead. 

@hacdias do you know how `snapStore` backend tells if this is official build or not?

PR #1171, opened 2019-10-05
feat: support experimental cohosting

hacdias:
Implements the initial spec of cohosting.

See https://github.com/ipfs-shipyard/cohosting.

This PR requires:

~- https://github.com/olizilla/ipfs-cohost/pull/5~

License: MIT

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1572,6 +1572,11 @@
       "resolved": "https://registry.npmjs.org/@hapi/file/-/file-1.0.0.tgz",
       "integrity": "sha512-Bsfp/+1Gyf70eGtnIgmScvrH8sSypO3TcK3Zf0QdHnzn/ACnAkI6KLtGACmNRPEzzIy+W7aJX5E+1fc9GwIABQ=="
     },
+    "@hapi/formula": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@hapi/formula/-/formula-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


hacdias:
@lidel probably I'm updating `https://github.com/olizilla/ipfs-cohost` to match our spec and I can just export the update function!

hacdias:
Please see https://github.com/olizilla/ipfs-cohost/pull/5. Let's wait for that

hacdias:
@lidel perhaps this would also be an easy way to apply the syncing part on IPFS Companion.

lidel:
Yes, I like the idea of reusing `ipfs-cohost` for sync/add/remove in ipfs-companion.

hacdias:
Pinging @lidel

lidel:
As agreed during last GUI/Browsers Weekly Call, let's pause this until 2020 Q1.
I don't think we should merge this without  solving https://github.com/ipfs-shipyard/ipfs-cohost/issues/6 first, and we have other priorities in Q4 (#1164).
Need to pick our  battles :)

bertrandfalguiere:
Will this feature be revived some day?

lidel:
Something like this will land, but additional UX work is needed. 
We need a clear user flow where act of pinning  a "website" or "page" via browser extension (https://github.com/ipfs-shipyard/ipfs-companion/issues/888) provides user with control over used storage (lazy vs full pin).

We won't be looking at  this  problem space until https://github.com/ipfs/ipfs-gui/issues/91 lands – I'm closing this for now.

PR #1195, opened 2019-10-13
fix: do not install npm-ipfs by default

### LINKED ISSUE #1194: `ipfs-npm` being installed automatically where it shouldn't
We couldn't make changes to your system. Please install or uninstall 'ipfs-npm' package manually

hacdias:
`ipfs-npm` was being install by default if the user had node installed for reasons that transcend my mere existence. I don't know why I wrote it that way, but certainly I shouldn't. This fixes that and closes #1194.

License: MIT

### DIFF
--- a/src/npm-on-ipfs/index.js
+++ b/src/npm-on-ipfs/index.js
@@ -1,5 +1,4 @@
 import which from 'which'
-import { ipcMain } from 'electron'
 import * as pkg from './package'
 import logger from '../common/logger'
 import store from '../common/store'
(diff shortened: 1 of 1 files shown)


src/npm-on-ipfs/index.js
```diff
@@ -30,27 +29,17 @@ export default function (ctx) {
   let opt = store.get(CONFIG_KEY, null)
   const exists = isPkgInstalled()
 
-  // Confirms if the package is still (un)installed because the user
-  // might change it manually.
-  if (opt !== null || exists !== false) {
+  if (opt === null) {
```
autonome: Is `exists` used anymore?

Can you leave the comment in, or correct to describe current behavior?

hacdias: Yes, `exists` is used, right on the next two lines. I did not write a comment because I thought the info provided by `logger.info` would work as a comment. If I add one, I will basically just copy the next line

Issue #1196, opened 2019-10-16
High CPU/GPU usage when ipfs daemon is running


rex4539:
ipfs-desktop 0.9.5

Steps:
Keep IPFS-desktop running in the background (menu item not open) with ipfs daemon running.

What happened:
High CPU/GPU usage. This forces laptop fans to work and eats battery like crazy.

<img width="1184" alt="Screenshot 2019-10-16 at 14 16 00" src="https://user-images.githubusercontent.com/227442/66914902-72f83680-f020-11e9-93d2-af7721d5a0d1.png">

Expected result:
Low CPU/GPU usage when the app is in the background (menu item not open).

Notes:
When ipfs daemon is stopped then the CPU/GPU usage is low (as it should). Is there a particular reason the app consumes so much CPU/GPU when the daemon is running (since the app window is NOT open)?

lidel:
Most likely (high GPU) suggests it is either Peers screen (world map) or Status (the canvas with `Bandwidth over time`) being rendered in the background:

@hacdias try creating an empty route in webui, and switch to it every time IPFS Desktop is minimized. This should remove the overhead of updating canvas when not visible.

PR #1198, opened 2019-10-18
fix: use blank page when window isn't active

hacdias:
This addresses #1196 by changing to a blank page whenever the window is not visible. For this, we use the Page Visibility API.

Requires https://github.com/ipfs-shipyard/ipfs-webui/pull/1259.

This reduced **significantly** the CPU usage on my end.

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "clean:webui": "shx rm -rf assets/webui/",
     "build": "run-s build:*",
     "build:webui": "run-s build:webui:*",
-    "build:webui:download": "npx ipfs-or-gateway -c QmeSXt32frzhvewLKwA1dePTSjkTfGVwTh55ZcsJxrCSnk -p assets/webui/ -t 360000 --verbose",
+    "build:webui:download": "npx ipfs-or-gateway -c QmRLGrLEjy1TsUBEyoPqSwdZwPjA48CjeZzJ7PD73nWyz4 -p assets/webui/ -t 360000 --verbose",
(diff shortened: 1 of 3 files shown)


autonome:
Looks good! Two questions:

* How does this impact performance of webui load and render?

* This feels a bit like giant hammer, without knowing more about the underlying cause of webui chewing cpu. Have you profiled why webui is doing that in the first place? Are there other issues tracking that?

hacdias:
> How does this impact performance of webui load and render?

It doesn't. We are always on the Web UI, but it's just an empty page.

> This feels a bit like giant hammer, without knowing more about the underlying cause of webui chewing cpu. Have you profiled why webui is doing that in the first place? Are there other issues tracking that?

The problem is that we were, by default, on status page. And when we're on the status page, we're continuously fetching repo and bandwidth stats to update the graphs. That creates load on the Web UI and `ipfs` itself.

autonome:
> The problem is that we were, by default, on status page. And when we're on the status page, we're continuously fetching repo and bandwidth stats to update the graphs. That creates load on the Web UI and `ipfs` itself.

Can you file a followup issue in webui repo for reducing the cost of having the status page open? You've reduced the cost on desktop in this PR, which is awesome. But if someone leaves the page open in a tab in the browser, they'll still have the problem.

hacdias:
@autonome done: https://github.com/ipfs-shipyard/ipfs-webui/issues/1260!

charlestonXdao (commented):
good

PR #1199, opened 2019-10-18
feat: set icon and category in Linux .desktop file

alexhenrie:
This patch gives IPFS Desktop the correct icon and puts it in the correct category in the desktop applications menu.


hacdias:
Thanks for this!

PR #1200, opened 2019-10-18
feat: improve auto launch at login

hacdias:
Auto Login was working... kinda of. According to the reports by @alexhenrie, auto launch on startup was not working on login since our installation path has a space on it and the [package](https://github.com/Teamwork/node-auto-launch/issues/99) we're using is not escaping that properly.

Unfortunately, [`node-auto-launch`](https://github.com/Teamwork/node-auto-launch/issues/99) hasn't been updated since 2017 and since then, Electron got its own capabilities to add login items for Windows and macOS so we can use that feature.

This changes the following:

- Stops using `node-auto-launch` for managing our launch at login state. It is still a required dependency so we can remove previous entries created by it. It will be removed on 0.10.0.
- Starts using `app.setLoginItemSettings` to manage login state on Windows and macOS.
- For Linux, we create a `.desktop` entry similar to the one `node-auto-launch` did, but with the execution path properly escaped.

@alexhenrie could you try this out on your end, please?

### DIFF
--- a/src/auto-launch.js
+++ b/src/auto-launch.js
@@ -18,9 +18,9 @@ function isSupported () {
 // Disable the old auto launch mechanism.
 // TODO: remove on 0.10.0.
 async function disableOldLogin () {
-  const autoLauncher = new AutoLaunch({ name: 'IPFS Desktop' })
-
   try {
+    const autoLauncher = new AutoLaunch({ name: 'IPFS Desktop' })
+
     if (await autoLauncher.isEnabled()) {
       await autoLauncher.disable()
       logger.error('[launch on startup] old mechanism disabled')
--- a/src/auto-launch.js
+++ b/src/auto-launch.js
@@ -42,8 +42,8 @@ async function enable () {
   const desktop = `[Desktop Entry]
 Type=Application
 Version=1.0
-Name="IPFS Desktop"
-Comment="IPFS Desktop Startup Script"
+Name=IPFS Desktop
+Comment=IPFS Desktop Startup Script
 Exec="${process.execPath}"
 StartupNotify=false
 Terminal=false`
--- a/src/auto-launch.js
+++ b/src/auto-launch.js
@@ -15,13 +15,24 @@ function isSupported () {
   return plat === 'linux' || plat === 'win32' || plat === 'darwin'
 }
 
+// Disable the old auto launch mechanism.
+// TODO: remove on 0.10.0.
+async function disableOldLogin () {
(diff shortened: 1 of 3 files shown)


alexhenrie:
I tested this code and it works for me. Thank you!

alexhenrie:
It would be nice though if you dropped the quotes from the `Name` line of the .desktop file, since those quotes do show up in the icon title when viewing `~/.config/autostart` in a file browser.

hacdias:
@alexhenrie does the comment also need to without quotes? I'm now wondering why's the path need to be quoted, while other things with spaces don't.

alexhenrie:
Only the path should be quoted. The quotes are needed there to distinguish the path to the binary from the arguments to the binary.

alexhenrie:
So yes, neither the `Name` line nor the `Comment` line should have quotes.

src/auto-launch.js
```diff
@@ -1,29 +1,86 @@
...
   const activate = async (value, oldValue) => {
     if (process.env.NODE_ENV === 'development') {
       logger.info('[launch on startup] unavailable during development')
       return
     }
 
+    if (!isSupported()) {
+      logger.info('[launch on startup] not supported on this platform')
+      return false
+    }
```
lidel: Are we hiding the "Launch on Startup" Option from WebUI's Settings when platform is not supported?

src/auto-launch.js
```diff
@@ -1,29 +1,86 @@
...
+}
+
+function getDesktopFile () {
+  return path.join(untildify('~/.config/autostart/'), 'ipfs-desktop.desktop')
+}
+
+async function enable () {
+  if (app.setLoginItemSettings) {
+    app.setLoginItemSettings({ openAtLogin: true })
+  }
```
lidel: To be sure there are no duplicated autoruns when `setLoginItemSettings` adds support for Linux, we could run this only on macos and windows, and finish immediately on those platforms, so useless `.desktop` is not created on them (same with `disable`):

```js
  if (app.setLoginItemSettings && ['win32', 'darwin'].includes(os.platform())) {
    app.setLoginItemSettings({ openAtLogin: true })
    return
  }
```

hacdias: Thanks. Completely forgot the `return`. However, I don't think adding the second verification on the `if` is necessary since the function only exists on supported platforms.

src/auto-launch.js
hacdias: Not sure, but will check.

Issue #1229, opened 2019-11-04
Change default install location for Win


papagunit:
Hi there,

It's not apparent, but is it possible to change the default installation location? I don't have anymore space on C and need it on a different drive.

hacdias:
@papagunit right now, it is not possible. It was in times, when our installer was a bit more complex.

Basically, we could revert to the "old style" installer which is not click once to install. You'd need to go through a few 'Next's but then you could choose the path.

Maybe useful: IPFS Desktop's installer is a NSIS installer. If there's any flag you could pass to a NSIS installer to change the installation path, then you might be able to do so with our installer. I'm not aware of any.

---

The biggest question here for the installer is: simplicity vs customization?

hacdias:
Possible now!

papagunit:
Christmas came early!

PR #1248, opened 2019-11-15
feat: use longer installer on Windows

hacdias:
This lets Windows users pick the location where they want to install IPFS Desktop and this addresses #1229. Bellow is the "next, next" journey Windows users are accustomed to:

Screenshot 2019-11-15 at 19 19 57
Screenshot 2019-11-15 at 19 20 02
Screenshot 2019-11-15 at 19 20 08
Screenshot 2019-11-15 at 19 20 19

^ We can change the picture that shows on the left. We should.


lidel:
@hacdias what are the dimension requirements of the image we can customize?
Perhaps we could reuse some of wallpapers made by @akrych?

hacdias:
@lidel here is what I've just pushed. It even aligns to macOS installer!

image

Issue #956, opened 2019-06-10
Tray Icon - Colors, more info?


hacdias:
Right now, IPFS Desktop tray icon is coloured. That goes against some guidelines and "what we usually" see on macOS. If we notice, when using the dark theme, we get white icons, and when using the light theme, we get dark/coloured icons.

The same applies for the most recent version of Windows 10, where there are effectively two themes (light and dark). System icons honour that, although since it is a really recent thing, most apps don't.

![image](https://user-images.githubusercontent.com/5447088/59181536-7570ee00-8b5f-11e9-8e64-09feb2013850.png)

![image](https://user-images.githubusercontent.com/5447088/59181875-33947780-8b60-11e9-9e60-234cca6dbbd7.png)

I think we could provide a coloured icon and a black and white version. By default, it would honour the system settings (dark or light theme) and we could let the user choose if they want a coloured icon or not in settings (like Mullvald does).

Also, we could expand our icon to show some interesting features. @daviddias has suggested, on https://github.com/ipfs-shipyard/ipfs-webui/pull/1048#issuecomment-500196126, the following:

> The most important question of all (:D), will my little IPFS Cube in my nav bar become the npm on IPFS cube?
> 
> <img width="159" alt="image" src="https://user-images.githubusercontent.com/1211152/59156911-2f565480-8aa3-11e9-9c5a-3ad1bbf09fb8.png">
> 
> <img width="183" alt="image" src="https://user-images.githubusercontent.com/1211152/59156913-35e4cc00-8aa3-11e9-9c60-ab272409eff0.png">
> 
> It would feel it gets equipped with a super power!

Despite our cube only having three faces, surely not most of the experiments/features deserve to be "shown off" on the icon. Some of them could be js-ipfs (#601) and npm-on-ipfs (#911).

/cc @ipfs-shipyard/gui @daviddias

lidel:
The same "dark" and "light" icon variants could be used in [ipfs-companion](https://github.com/ipfs/ipfs-companion/) in Firefox
(it supports [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action))

lidel:
The same "dark" and "light" icon variants could be reused in [ipfs-companion](https://github.com/ipfs/ipfs-companion/) in Firefox
[...]

PR #1256, opened 2019-11-22
feat: apple guidelines macOS compliant icon

hacdias:
After much discussion on #956, I decided to keep it simple and just change the menubar icon on macOS since it is the only platform where we are actually violating the guidelines.

Secondly, after a lot of discussion (again!) about how to differentiate between an online and offline node, I decided to go with the most simple solution and follow what some other tools do: just dim the icon when it's offline

@ericronne I used your last icon. However, I needed to **resize** it because it had a lot of white space surrounding it. Since it's a real pixel perfect work, I ask you to take a look at this screenshots to see if they look good or not, please.

Screenshot 2019-11-22 at 06 13 41
Screenshot 2019-11-22 at 06 13 36
Screenshot 2019-11-22 at 06 13 29
Screenshot 2019-11-22 at 06 13 21


lidel:
If we use dimming for status, do we still need dots in corners?
Perhaps a simple cube would be enough.

hacdias:
@lidel it works for me. Wdyt @ericronne ?

ericronne:
Sure, thx!

hacdias:
@ericronne could you make a small cube icon with stronger outlines so we can apply it here, please?

ericronne:
Here's both a plain option one with node dots on the corners. The plain version better represents the most widely used ipfs logo, but it's also more generic ...
image
**[plain .svg](https://drive.google.com/open?id=1xqcFvaeuPmZoQy1Zm2tJhzFkTQnIxY3Z)**

I'd recommend the dotted version …
image
**[dotted .svg](https://drive.google.com/open?id=1VWG9c3-U6dnGBgVr1u1UxZbxWqvoYXpz)**

hacdias:
Let's vote!

:tada: for

image

:rocket: for

image

hacdias:
PS: our icon is perfectly centered. Other might not and cause the sensation of not being centered.

autonome:
In isolation and enlarged, the dotted version looks ok. However, when in the menubar it no longer looks easily recognizable as IPFS - the large dots make it look circular, a bit like a ship's wheel.

I recommend using either the generic-but-more-obviously-IPFS icon or experimenting with smaller dots.

hacdias:
Say hello to the winner and to the next "generation" IPFS Desktop on macOS icon:

image

PR #1270, opened 2019-11-28
fix: ignore net:: errors

### LINKED ISSUE #1267: net:: crashes
This is a issue to find out why there are `net::` crashes. Maybe we can just handle that case specifically on unhandled rejections handler not to throw anything and just be ignored.

hacdias:
This aims to fix #1267. I don't know how to reproduce the error so I'm not sure if this is going to fix it. Those errors never have stack trace so it's hard to understand them. However, I believe this might fix it because it just ignores errors with `net::` on it generated by unhandled rejections and exceptions.

License: MIT

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ app.on('will-finish-launching', () => {
 function handleError (err) {
   // Ignore network errors that might happen during the
   // execution.
-  if (err.toString().includes('net::')) {
+  if (err.stack.includes('net::')) {
     return
   }
 
--- a/src/index.js
+++ b/src/index.js
@@ -40,6 +40,12 @@ app.on('will-finish-launching', () => {
 })
 
 function handleError (err) {
+  // Ignore network errors that might happen during the
+  // execution.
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -40,6 +40,12 @@ app.on('will-finish-launching', () => {
 })
 
 function handleError (err) {
+  // Ignore network errors that might happen during the
+  // execution.
+  if (err.stack.includes('net::')) {
+    return
+  }
+
   logger.error(err)
```
lidel: could we move this logger before the `if`, so its always logged?

hacdias: I was just thinking about ignoring the noise caused by Internet issues...

lidel: Just to be sure I got this right, this is when ipfs-desktop tries to access API that is offline, or are there other uses of this code path?

hacdias: AFAIK, it's when we try to access the API and it is offline. However, it shouldn't happen many times.

On previous occasions, this would happen with Auto Updater, when we didn't catch the errors.

PR #1302, opened 2019-12-22
feat: new ctl

hacdias:
Closes #1137

@lidel I'm pretty sure you're more aware of the changes made to this version than I am. Could you give me a hand with this? Also, I noticed `ipfsd.stop` hangs and it does not take timeouts anymore...


lidel:
Will look into this, but it requires bigger refactor than I thought.
In case  I don't post any updates later today, we will get back to this after holidays.

src/daemon/daemon.js
```diff
@@ -43,18 +41,28 @@ async function cleanup (ipfsd) {
...
-    repoPath: path,
+    test: false,
+    args: flags,
+    type: type
+  })
+
+  const ipfsd = await factory.spawn({
+    ipfsOptions: {
+      repo: path
+    },
```
lidel: I believe basic init did not work because repo path got moved to `ipfsOptions`.  This was the fix :)

hacdias: About the `api` file: I thought IPFS itself handled it! The only thing we did was this: #722.

hacdias:
I'm taking a new look at this on #1411.

hacdias:
I'm closing this in favour of #1411. Thanks for all your work here @lidel. I've reused most of it ;)

PR #1360, opened 2020-02-24
chore: helper scripts for updating chocolatey

hacdias:
This PR creates an helper script to help with the chocolatey update procedure. After releasing a version, wait for the binaries to be available on the releases page. Then:

1. `node pkgs/chocolatey/update.js $version`
2. `git commit -m "chore: update choco [pubchoco]"` - the `pubchoco` part *must be there*
3. `git push` and wait for AppVeyor which should publish the package.

Obviously I could not test this. The script for updating the nuspec file works well. We just need to test the third step on the next release.


appveyor.yml
```diff
@@ -22,4 +22,14 @@ build_script:
   - npm run test:e2e
   - npm run build
 
+on_success:
+  - ps: |
+      if ($env:APPVEYOR_REPO_COMMIT_MESSAGE -Match "pubchoco") {
```
lidel: > Ugh, github lost my comment, so writing again...

Appveyor runs on every PR. In theory someone could PR a downgrade attack with `nuspec` pointing at older version.  What would Cocolatey do in that case? Reject? Override latest version with old one?

hacdias: Changed. However, couldn't someone just change the code? And mame a PR and hijack the releases anyways? There's a way to define scripts that does not include `.appveyor.yml` (see [docs](https://www.appveyor.com/docs/build-configuration/)). However, someone with enough knowledge could trigger the update too...

hacdias: Actually not, if we define like they do on the docs (on the UI itself), it won't be on `.appveyor.yml` thus it will be locked to `master`.

hacdias:
@lidel so I removed that bit from .appveyor.yml and added it to the interface:

image

This way no one can change it!

lidel:
@hacdias I tried `pubchoco` in https://github.com/ipfs-shipyard/ipfs-desktop/commit/c23ff544a9811bfdb536a319e91a55ae7f037ac7  in but did not seem to publish new version at https://chocolatey.org/packages/ipfs-desktop 

I know you don't have bandwidth for looking into this, so let's leave automation for now.
Do you mind publishing 0.10.4 manually, in spare time?

hacdias:
@lidel done. I'm sad it didn't work. Looking at the logs, it doesn't even seem that the script was executed...

Issue #1211, opened 2019-10-23
'IPFS Desktop' can't be opened because Apple cannot check it for malicious software


brennanfife:
Reinstalled Brave with the IPFS Chrome extension and the IPFS Companion Desktop application this morning. Opened it, but only get this warning with the following description: 'this software needs to be updated. Contact the developer for more information.' 

Currently running macOS Catalina 10.15

pcowgill:
FYI this still is an issue for fresh installs

andrew:
I ran into the same error this morning, it can be fixed by IPFS devs "Notarizing" the app with Apple before releasing it: https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution

lidel:
Reopening. We need someone with Mac for this:

> Notarization **requires Xcode 10** or later. 
> Building a new app for notarization **requires macOS** 10.13.6 or later.

hacdias:
Ref.: https://github.com/karaggeorge/electron-builder-notarize

lidel:
No Mac at hands, but I moved the needle a bit:

- switched (https://github.com/ipfs-shipyard/ipfs-desktop/pull/1341) to the latest electron-builder to include fixes needed for notarization of mac artifact 
  - see "mac" changes [here](https://github.com/electron-userland/electron-builder/releases/tag/v22.3.1)
   - and [electron-notarize's prerequisites](https://github.com/electron/electron-notarize#prerequisites)

lidel:
For now, I think we could try to manually notarize v0.10.3

Ref.
- IIUC we need someone with Xcode and AppleID
  - I am not sure if it needs to be the same ID as one used for signing:
    ```
    signing         file=dist/mac/IPFS Desktop.app identityName=Developer ID Application: Protocol Labs, Inc. (7Y229E2YRL) identityHash=C297E064280322D72444C869E7FC0F60B14E934E provisioningProfile=none
    ```
    @olizilla do you remember who owns keys used at Travis by any chance?
- Try executing step (2) described at [_Notarize Your Preexisting Software_](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution#3087722), but note that:
  > Alternatively, you can put apps, kernel extensions, and other software in a container, like a disk image, and notarize the container. The notary service accepts disk images (UDIF format), signed flat installer packages, and ZIP archives.
[...]

Issue #66, opened 2015-10-19
Binary signing for macOS


dignifiedquire:

Dignifiedquire:
@jbenet how long do you think this will take you? Is it realistic to get this done for 1.0?

jbenet:
i'm not sure. i'll look into it, but this week is basically shot for me. if you need it this week, then no. next week is likely. then there's apple's review process. i think it's much faster now, but it used to take a week or something.

Dignifiedquire:
That's fine, I'll do the prerelease unsigned this week and want to wait at least one week anyway before the real release.

Dignifiedquire:
Ref https://github.com/maxogden/electron-packager/issues/163

olizilla:
Happy to say we're getting a first pass on "signed releases for macOS" in to the next release, in significantly less than 4 years after the initial proposal.

Of note, it is tricky. There are a bunch of hoops you have to jump through on the Apple Developer portal and more on the local keychain handling, and more again to wire it up to CI.

- You need to be the Apple Developer "Team Agent" to create certificates. 
- There can be only one Team Agent,  having the admin role wont do.
- You must have 2 factor auth enabled on your apple ID.
- You must create certificate signing request via the Keychain Access app to create certificates.
- You must create both a `Developer Installer` and a `Developer Application` certifcate, and download them to your local keychain.
- You have to access them via the "My Certificates" filter, to enable the "export as .p12" option. If you navigate to them via another filter or search, that export option is not available. You better be the team agent if you want that to work. If all the planets are aligned you should be able to export them as a combined, encrypted `.p12` file like so

![add-apple-certs](https://user-images.githubusercontent.com/58871/51535890-b0d57f00-1e41-11e9-8f2d-b2d959016e25.gif)

Then follow the steps in https://www.electron.build/code-signing#travis-appveyor-and-other-ci-servers to wire it up for CI

> To sign app on build server you need to set CSC_LINK, CSC_KEY_PASSWORD:
- Export certificate. Consider to not use special characters (for bash) in the password because “values are not escaped when your builds are executed”.
- Encode file to base64 (macOS: `base64 -i yourFile.p12 -o envValue.txt`, Linux: `base64 yourFile.p12 > envValue.txt`).

Thanks to @jesseclay for sticking with me on this advenure!

olizilla:
I figured this out from a mix of
- https://www.electron.build/code-signing
[...]

PR #1365, opened 2020-02-27
chore: macOS notarizing

lidel:
## Motivation 
This  PR aims to fix macOS Catalina issues described in  https://github.com/ipfs-shipyard/ipfs-desktop/issues/1211

### Details

This adds scripts that run [electron-notarize](https://github.com/electron/electron-notarize) as additional manual or build steps on darwin runtime, loosely following https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/

#### `pkgs/macos/notarize-build.js`  (not used for now)

This runs as part of `darwin` build as it was suggested on https://www.electron.build/code-signing.

Skipped if `process.env.APPLEID` or `process.env.APPLEIDPASS`  are not set at the build time. As there was no safe way to do it on TravisCI, we don't use it at CI atm.

#### `pkgs/macos/notarize-cli.js` (should do for now)

A standalone CLI tool that performs notarization of existing macOS artifact.
It enables us to notarize macOS artifact as a semi-automatic step during release dance:

1. Download .dmg from https://github.com/ipfs-shipyard/ipfs-desktop/releases/vA.B.C
2. Run `node pkgs/macos/notarize-cli.js ./path/to/IPFS-Desktop-A.B.C.dmg`


pkgs/macos/notarize.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const appName = context.packager.appInfo.productFilename
+
+  return notarize({
+    appBundleId: 'io.ipfs.desktop',
+    appPath: `${appOutDir}/${appName}.app`,
+    // TODO: figure out how to get credentials ( https://github.com/electron/electron-notarize#method-notarizeopts-promisevoid)
+    //        Q: use user & pass (below) or appleApiKey & appleApiIssuer?
+    appleId: process.env.APPLEID, // TODO: set this on CI?
+    appleIdPassword: process.env.APPLEIDPASS // TODO: set this on CI? it needs to be app-specific, generated on https://appleid.apple.com/
```
lidel: We need to figure out this part next.

olizilla:
Relevant context for the macOS binary signing that we set up previously https://github.com/ipfs-shipyard/ipfs-desktop/issues/66#issuecomment-456384158

lidel:
@autonome 
I've discussed this with @olizilla and threat model is tricky. 
Doing CI-based notarization will require us to move away from TravisCI.  

We have no bandwidth nor full time maintainer atm. Let's  go with semi-automatic notarization during release dance.

Created a script for post-build online notarization (without stapling) – should be enough for now:

```console
$ node pkgs/macos/notarize-cli.js ./path/to/IPFS-Desktop-A.B.C.dmg
```

(People download `.dmg`, so are online anyway)

README.md
```diff
@@ -119,6 +119,9 @@ Other languages are periodically pulled from [Transifex](https://www.transifex.c
 - Publish local changes and the tag to GitHub repo: `git push && git push origin vA.B.C`
 - Wait for the CI to upload the binaries to the draft release (a new one will be created if you haven't drafted one).
 - The `latest.yml, latest-mac.yml, latest-linux.yml` files on the release are used by the app to determine when an app update is available. Once a release is published, users should recieve the app update. See: https://www.electron.build/auto-update.
+- Notarize `.dmg` at Apple (context: [#1365](https://github.com/ipfs-shipyard/ipfs-desktop/issues/1211))
+    1. Download `.dmg` from `https://github.com/ipfs-shipyard/ipfs-desktop/releases/vA.B.C`
+    2. Run `node pkgs/macos/notarize-cli.js ./IPFS-Desktop-A.B.C.dmg`
```
autonome: What do you do w/ the notarized dmg? Commit back to repo?

autonome: Should adding and editing the .env file be here too?

lidel: #### what to do with notarized .dmg?

We do nothing with notarized `.dmg`, just discard it for now. Goal is to inform apple that this is a legit binary blob, so it can give thumbs up when someone's macOS asks about it in the future.

Why? **We can't replace `.dmg` attached to GitHub Release with notarized version** because checksum in `latest-mac.yml` would no longer match and that would break autoupdates. 

But online notarization should still work: when macOS sees a new version of ipfs-desktop it will check for stapled notarization, won't find it, so it will fallback to asking Apple servers if running it is ok. Autoupdate requires internet connection anyway, so this non-stapled notarization should also work.

#### env

`notarize-cli` will inform user if env variables are missing

lidel:
```suggestion
    2. Ensure `APPLEID` and `APPLEIDPASS` are set either as environment variables or entries in `.env` file. Those need to belong to the same org as cert used for signing.
    3. Run `node pkgs/macos/notarize-cli.js ./IPFS-Desktop-A.B.C.dmg`
```

olizilla:
worth noting that we are depending on `electron-notarize-dmg` which is a user provided fork of `electron-notify`... it's pinned at 1.0.0, which is good as we don't trust any old updates of this fork... the github diff for the fork seem legit, but thats no guarantee of whats inside the published module https://github.com/electron/electron-notarize/compare/master...hugozap:master

lidel:
@olizilla yes, I pinned it for that very reason. For what its worth I checked code from NPM in `node_modules` and the diff is matching github one.

(Manual cli notarization is a temporary solution until we get a full time maintainer who will have bandwidth for doing in on the CI safely)

olizilla:
The underside of this iceberg required us to also

- upgrade locally installed xcode to the latest version (11.3)
- generate a new app-specific password in the apple developer portal
- renew our apple developer license, for $99 per year.
- sign agreements for unpaid and, suprisingly, paid apps even though we only want to notarize a free app that we do not distribute through the app store.
  - many thanks to https://forums.developer.apple.com/message/397892#397892

with all that in place, running 
```
node pkgs/macos/notarize-cli.js ~/Downloads/IPFS-Desktop-0.10.4.dmg
```
returned without error. The overlord blessed us with an email.

Screenshot 2020-03-12 at 10 51 14

olizilla:
Can confirm... on macos 10.15.3, trying to run ipfs-desktop 0.10.3 results in

Screenshot 2020-03-12 at 14 18 07

while trying trying to run 0.10.4 now gives us a

Screenshot 2020-03-12 at 14 18 30

`\o/`

olizilla:
> For what its worth I checked code from NPM in node_modules and the diff is matching github one.

very cool @lidel. did i mention that you are the best?

lidel:
@olizilla no, you are! thank you for pushing this through the finish line :heart:

Issue #1153, opened 2019-09-27
Tray Icon problems on Linux


hacdias:
So, there are quite a few issues with people complaining they can't access the UI of IPFS Desktop on Linux. All of them are caused by the same reason: tray icons are being deprecated from many desktop environments and Electron has not yet worked out on a solution, nor it seems to be [happening soon](https://github.com/electron/electron/issues/10427).This affects all kinds of packages.

One possible way for us to circumvent this for now is to open the main Web UI window on Linux on startup. However, that will still leave the options of exiting desktop and other options provided by the menu behind.

I'm opening this in hope to gather feedback on how to create an workaround around this.

Please note that Linux support is **experimental** nonetheless.

NatoBoram:
Hm, not a fan of this solution.

Tray icons are an important feature. The desktops missing them are just deficient, intentionally or not, and regardless of their design philosophy.

A better solution would be to open a window when clicking on the application's launcher. That way, it now gains a behavior that's intuitive (clicking on a launcher generally launches the app) and it fixes the tray icon problem rather than creating a missing tray icon problem for other Linux users.

olizilla:
As a first step, we can have desktop open the webui window on launch as suggested. I'm not clear on how we can show the user that the app is still running if they close the window... what happens on others apps that are long running on desktops that dont have task tray icons?

NatoBoram:
> what happens on others apps that are long running on desktops that dont have task tray icons?

I don't know any long-running apps without tray icons except for command-line apps, but those don't even have desktop icons in the first place. For long-running apps *with* a tray icon on systems *without* a system tray, however, I've noticed a pattern.

* [Skype](https://www.skype.com/) launches its UI when clicking its icon even if it's already launched in background.
* [Discord](https://www.skype.com/) launches its UI when clicking its icon even if it's already launched in background.
* [Steam](https://store.steampowered.com/) launches its UI when clicking its icon even if it's already launched in background.
* [Chrome](https://www.google.com/chrome/) launches its UI when clicking its icon even if it's already launched in background.
* IPFS-Desktop doesn't react.

None of them closes on close. They minimize on minimize and go to system tray on close.

One problem that may arise is that options in the tray app will be hidden. This can be fixed by having them accessible from inside the app as well.

hacdias:
> ...we can have desktop open the webui window on launch as suggested.

[...]

PR #1370, opened 2020-03-10
fix: startup on Debian 10

### LINKED ISSUE #1362: ipfs-desktop > Trace/breakpoint trap
installed from latest .deb on a debian derivate

on launch
$ipfs-desktop 
$Trace/breakpoint trap

lidel:
> Closes #1362 

This PR fixes startup on Debian 10 and any other distribution with strict / b0rked Chromium sandbox setup.

We set `executableArgs` to `--no-sandbox` in `electron-builder.yml`.
It will update `Exec` command in `ipfs-desktop.desktop` manifest that is run when menu item / icon in desktop environment is clicked. 

### Motivation

App does not start on Debian 10 due to sandboxing FUBAR.
Details in https://github.com/ipfs-shipyard/ipfs-desktop/issues/1362#issuecomment-596857282)

Note: this should be acceptable for this app.
Electron is used only for tray (ha, #1153) menu, the GUI for starting/stopping daemon. Actual IPFS node is a golang binary running in separate process.


package.json
```diff
@@ -58,8 +58,8 @@
     "cross-env": "^6.0.3",
     "delay": "^4.3.0",
     "dirty-chai": "^2.0.1",
-    "electron": "^7.1.11",
-    "electron-builder": "^22.3.2",
+    "electron": "^7.1.14",
```
lidel: I did not update to 8.x because it does not support old school Tray icons on Linux (#1153) and properly addressing that requires time.

autonome: Screen Shot 2020-03-10 at 10 18 45 AM

README.md
```diff
@@ -164,6 +165,31 @@ Since we're using [ipfsd-ctl](https://github.com/ipfs/js-ipfsd-ctl), we have our
...
+
+or a very short one, when starting in a terminal:
+
+```console
+$ ipfs-desktop
+$Trace/breakpoint trap
+```
+
+
+**TL;DR** There are two ways of fixing this, pick one that works for you:
```
autonome: Seems like the first option opens users up to more issues. Any reason why we wouldn't recommend only the solution specific to IPFS?

lidel: After giving this more thought I removed `kernel.unprivileged_userns_clone` one.
It is not present in regular Linux kernel (Debian added it as part of their hardening patches) and won't work for most people anyway.

README.md
```diff
@@ -164,6 +165,33 @@ Since we're using [ipfsd-ctl](https://github.com/ipfs/js-ipfsd-ctl), we have our
...
+```
+
+This is a known issue with Electron/Chrome and some hardened kernels.
+If you are interested in details, read [this](https://github.com/ipfs-shipyard/ipfs-desktop/issues/1362#issuecomment-596857282).
+
+The only reliable way to fix this at the moment is to start the app with additional parameter:
+
+```console
+$ ipfs-desktop --no-sandbox
+```
```
lidel: Updated troubleshooting section contains a single, clear advice.

lidel:
Merging to unblock release, we can tweak troubleshooting in README in separate PR

PR #1385, opened 2020-03-26
fix: call stopIpfs on ctx

### LINKED ISSUE #1375: stopIpfs is not a function
I clicked update IPFS and an error window appeared.

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.10.3
- **Electron Version**: 7.1.11
- **Chrome Version**: 78.0.3904.130

**Error**

```
TypeError: stopIpfs is not a function
    at _callee$ (D:\network\IPFS\resources\app.asar\out\auto-updater\quit-and-install.js:42:20)
    at tryCatch (D:\network\IPFS\resources\app.asar\node_modules\regenerator-runtime\runtime.js:45:40)
    at Generator.invoke [as _invoke] (D:\network\IPFS\resources\app.asar\node_modules\regenerator-runtime\runtime.js:271:22)
    at Generator.prototype.<computed> [as next] (D:\network\IPFS\resources\app.asar\node_modules\regenerator-runtime\runtime.js:97:21)
    at asyncGeneratorStep (D:\network\IPFS\resources\app.asar\node_modules\@babel\runtime\helpers\asyncToGenerator.js:3:24)
    at _next (D:\network\IPFS\resources\app.asar\node_modules\@babel\runtime\helpers\asyncToGenerator.js:25:9)
    at D:\network\IPFS\resources\app.asar\node_modules\@babel\runtime\helpers\asyncToGenerator.js:32:7
    at new Promise (<anonymous>)
    at D:\network\IPFS\resources\app.asar\node_modules\@babel\runtime\helpers\asyncToGenerator.js:21:12
    at _quitAndInstall (D:\network\IPFS\resources\app.asar\out\auto-updater\quit-and-install.js:58:26)
```

hacdias:
This fixes #1375 by calling `stopIpfs` on `ctx`. I think that deconstructing `ctx` may be causing that error if and only if `quitAndInstall` was being called before `setupDaemon` [finished executing](https://github.com/ipfs-shipyard/ipfs-desktop/blob/master/src/index.js#L72).

Please validate if this makes sense @lidel.

License: MIT

### DIFF
--- a/src/auto-updater/quit-and-install.js
+++ b/src/auto-updater/quit-and-install.js
@@ -3,14 +3,14 @@ import { autoUpdater } from 'electron-updater'
 import { STATUS } from '../daemon'
 
 // adapted from https://github.com/electron-userland/electron-builder/issues/1604#issuecomment-372091881
-export default async function quitAndInstall ({ stopIpfs }) {
+export default async function quitAndInstall (ctx) {
   app.removeAllListeners('window-all-closed')
   const browserWindows = BrowserWindow.getAllWindows()
   browserWindows.forEach(function (browserWindow) {
     browserWindow.removeAllListeners('close')
   })
 
-  const status = await stopIpfs()
+  const status = await ctx.stopIpfs()
 
   if (status === STATUS.STOPPING_FAILED || status === STATUS.STOPPING_FINISHED) {
     autoUpdater.quitAndInstall(true, true)


hacdias:
@lidel `ctx.stopIpfs` seems to be a function:

```
 const runAndStatus = (fn) => async () => { 
   await fn() 
   return status 
 } 
```

Calling `runAndStatus` returns a function and it does not execute what's inside unless it's called again.

hacdias:
@lidel I agree with you that this is not the root cause of that problem. It doesn't make sense and the user must be really fast (probably impossible) to make this happen. I'll close this and see if I can reproduce.

PR #1392, opened 2020-04-07
feat: support for go-ipfs 0.5

lidel:
This bumps go-ipfs version to 0.5.0-rc1 in preparation for 0.5 release.

Small set of fixes was required:
- Status check
  - switched API calls from `GET` to `POST` (https://github.com/ipfs/go-ipfs/pull/7097)
- Bundled ipfs-webui loaded via Electron's BrowserWindow
  - `Origin` header: a proper value is set instead of removing it
  - sending `ipfs-desktop/${version}` in `User-Agent` header

### TODO before merge

- [ ] ~switch from `"v0.5.0-rc1` to `"v0.5.0` when it ships~ (new PR, see https://github.com/ipfs-shipyard/ipfs-desktop/pull/1392#issuecomment-618093160)

cc @hacdias @Stebalien  https://github.com/ipfs/go-ipfs/issues/7109

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ init:
   - git config --global core.autocrlf input
 
 install:
-  - ps: Install-Product node 10 x64
+  - ps: Install-Product node 12 x64
   - npm install
 
 build_script:
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "start": "cross-env NODE_ENV=development electron src/index.js",
     "lint": "standard",
     "test": "cross-env NODE_ENV=test mocha test/unit/**/*.spec.js",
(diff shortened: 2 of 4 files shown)


lidel:
Old way of dealing with CORS no longer worked, I replaced it with a proper `Origin` that looks like request comes from a page loaded from API origin:

webui-desktop-2020-04-07--20-59-44

lidel:
Rebased on master, switched to 0.5.0-rc3.
Should be ready to merge (we can switch to 0.5.0 when it is released)

lidel:
Ok, I think we need to merge https://github.com/ipfs-shipyard/ipfs-desktop/pull/1411 first – E2E tests won't pass without fixes present there.

PR #1407, opened 2020-04-17
feat: run garbage collector

### LINKED ISSUE #1355: Add manual garbage collector button
Hi!

I was looking to manually start the garbage collector but did not find any button to do it. I think it would be useful to have an option to do so.

This issue shows that it has been implemented in the past: https://github.com/ipfs-shipyard/ipfs-desktop/issues/635 (closed by https://github.com/ipfs-shipyard/ipfs-desktop/pull/636)

I can't find when it was scrapped, though.

Have a nice day!

hacdias:
Supports running the GC from the menubar. This was a feature that was lost during the revamp. Closes #1355.

Screenshot 2020-04-17 at 10 25 17

Screenshot 2020-04-17 at 10 25 22

Screenshot 2020-04-17 at 10 25 34

Note: the Electron icon is replaced by IPFS Desktop's icon on the final version.

License: MIT

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -43,6 +43,7 @@
   "takeScreenshot": "Take Screenshot",
   "downloadHash": "Download Hash",
   "moveRepositoryLocation": "Move Repository Location",
+  "runGarbageCollector": "Run Garbage Collector",
   "polkitDialog": {
     "title": "Polkit not found",
     "message": "IPFS can't be added to /usr/local/bin/ without polkit agent."
@@ -131,5 +132,14 @@
   "updateDownloadedNotification": {
     "title": "Update downloaded",
     "message": "Update for version { version } of IPFS Desktop downloaded. Click this notification to install."
+  },
+  "runGarbageCollectorWarning": {
+    "title": "Garbage collector",
+    "message": "Running the garbage collector will remove all the objects that are not pinned or on your MFS from your repository. This action might take a while. Do you want to run the Garbage Collector?",
+    "action": "Run GC"
+  },
+  "runGarbageCollectorDone": {
+    "title": "Garbage collector",
+    "message": "Garbage collector run was successful."
   }
 }
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -135,11 +135,11 @@
   },
   "runGarbageCollectorWarning": {
     "title": "Garbage collector",
-    "message": "Running the garbage collector will remove all the objects that are not pinned or on your MFS from your repository. This action might take a while. Do you want to run the Garbage Collector?",
-    "action": "Run GC"
+    "message": "Running the garbage collector will remove from your repository all objects that are not pinned or on your MFS. This may take a while. Do you wish to proceed?",
+    "action": "Run"
   },
   "runGarbageCollectorDone": {
     "title": "Garbage collector",
-    "message": "Garbage collector run was successful."
+    "message": "The garbage collector ran successfully."
   }
 }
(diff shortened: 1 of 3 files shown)


assets/locales/en.json
```diff
@@ -116,5 +117,14 @@
   "itemsFailedNotification": {
     "title": "Failed to add items",
     "message": "Could not add your items to your node."
+  },
+  "runGarbageCollectorWarning": {
+    "title": "Garbage collector",
+    "message": "Running the garbage collector will remove all the objects that are not pinned or on your MFS from your repository. This action might take a while. Do you want to run the Garbage Collector?",
+    "action": "Run GC"
```
jessicaschilling, line 124: Maybe just “Run” - dialog is clear enough and having to parse what “GC” means can be confusing

assets/locales/en.json
```diff
@@ -116,5 +117,14 @@
   "itemsFailedNotification": {
     "title": "Failed to add items",
     "message": "Could not add your items to your node."
+  },
+  "runGarbageCollectorWarning": {
+    "title": "Garbage collector",
+    "message": "Running the garbage collector will remove all the objects that are not pinned or on your MFS from your repository. This action might take a while. Do you want to run the Garbage Collector?",
```
jessicaschilling, line 123: Maybe:
“Running the garbage collector will remove from your repository all objects that are not pinned or on your MFS. This may take a while. Do you wish to proceed?”

assets/locales/en.json
```diff
@@ -116,5 +117,14 @@
...
+    "action": "Run GC"
+  },
+  "runGarbageCollectorDone": {
+    "title": "Garbage collector",
+    "message": "Garbage collector run was successful."
```
jessicaschilling, line 128: Maybe: “The garbage collector ran successfully.”

assets/locales/en.json
hacdias: Yes, just "Run" is much better! :D

assets/locales/en.json
hacdias: Perfect!

hacdias:
Thanks for the copy feedback @jessicaschilling!

> Will it ever NOT run successfully? If so, how does the user know? If not, maybe the languages should just be “garbage collection run finished” or similar.

I believe it's unlikely, but it may happen. And yes, the user will end up knowing. We'll tell them that an error occurred and ask them if they'd like to open an issue here with the error report (which is generated automatically!)

> I assume it runs in the background so user can still work during “it may take a while”?

Yes!

jessicaschilling:
Awesome, thanks for clarifying!

src/run-gc.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    const ipfsd = await getIpfsd()
+
+    if (!ipfsd) {
+      return
+    }
+
+    try {
+      ipfsd.api.repo.gc()
```
lidel, line 33: I know this is still for the old API, but when you switch to the new one in #1411 `repo.gc` will return async iterable which yields an Error or one line each time:
https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/REPO.md#repogc

I think the proper blocking and error handling would be something like:

```js
let error // keep first error, or this could be array too
for await (const res of ipfs.repo.gc()) {
  if (typeof res === 'Error') error = res
}
if (error) throw error
```

@hacdias  mind updating #1411 to  and fixing in a separate PR, if needed?

hacdias: Thanks for the heads up! Wasn't aware of the change on that API.

RubenKelevra:
Is there a progress bar or a wait icon of some kind? Since the operation can take extremely long if the datastore is large or the disk is slow.

hacdias:
No, but I was thinking of changing the tooltip while it's running so the user knew:

image

What do you think?

hacdias:
Uh, I have an idea with better UX. Will get back with a PR.

Issue #1389, opened 2020-04-03
[meta] decoupling from ipfs-webui


lidel:
> Just writing down some thoughts and concerns, needs more analysis, probably can be split into 2-3 phases. 

Right now ipfs-desktop is tightly coupled with ipfs-webui:
1. ships with a bundled version of ipfs-webui
2. opens bundled webui in Electron's webview
3. injects own settings into Settings screen of ipfs-webui

I believe it comes at significant maintenance cost and user confustion:

- requires ipfs-webui to support Electron "browser"
- every time new setting is added, ipfs-desktop release is blocked by ipfs-webui
- when user opens webui from commandline or via browser extension they see the same ipfs-webui, but when they go to Settings ipfs-desktop options  are not there (they are there only if webui is opened via ipfs-desktop menubar)

I suspect we could consider:

- remove Desktop-specific items from Settings screen in webui, and put them in menubar submenu instead
- opening webui from `http://127.0.0.1:8080/webui` instead of `webui://-` pseudoprotocol
  - to ensure webui is present in repo and works out of the box, bundle webui as a TAR archive with ipfs-desktop and import it if it is not not in local repo already
    (we have prior art of doing just that in Brave)
- opening in real web browser instead of Electron

jessicaschilling:
Is there any reason for the Electron standalone app to exist other than the fact that it comes as part of a convenient package that also installs the daemon and CLI tools?

lidel:
I believe _raison d'être` is to provide GUI for starting/stopping IPFS node and opening webui for everything else:
![](https://user-images.githubusercontent.com/157609/55424318-426b1680-5580-11e9-93ec-ec261879367f.jpg)

jessicaschilling:
Right. All of which could stand on their own and send the user to webUI as needed ...

hacdias:
Hey y'all!

> remove Desktop-specific items from Settings screen in webui, and put them in menubar submenu instead

Personally, I agree. Most of the options (if not all), are just checkboxes and Electron's menus have support for check-boxes type of things.

Also, by removing Web UI screen, we would simplify the code a lot and just need to open Web UI on user's default browser.

> opening in real web browser instead of Electron

I also agree with this. It simplifies a ton of things. There are a few things that'd need to be changed in order for all of this to work, but it would be quite simple.

---

[...]

PR #1413, opened 2020-04-17
feat(un): decouple webui [META] [THOUGHT]

hacdias:
See #1389.

This is my attempt at decoupling IPFS Desktop from Web UI. It is not finished and I won't work more on it unless we all agree we should move forward on this.

This basically removes the Web UI window and when you click 'Status' or 'Files' you're just pointed to your default browser, opening the Web UI.

image

The Preferences (or Settings for non-macOS users) are all checkboxes. Clicking on them should be enough to enable or disable them. Some (not implemented) should show a dialog explaining what they actually do and ask the user if they really want to proceed.

On the Web UI side, a lot of simplifications can be made with this!


hacdias:
@lidel @jessicaschilling @autonome I would love your input on this 😃

jessicaschilling:
Thought: Regardless of the degree to which we do/don't kill off the Electron standalone window in favor of moving into the browser -- moving those desktop-only settings out of the "Settings" screen and into the menubar makes a lot of sense. There's some linguistic stuff that probably needs to be refined (basically, we need to be explicitly clear about when "settings" means "IPFS config" and when it means "app settings"), but overall this makes a ton of sense, particularly for people who switch between Desktop and WebUI: right now we are giving them two different settings screens and that makes it really hard to remember where you do what.

If we do go in the direction of "IPFS Desktop" really just meaning "IPFS Menubar App", this sort of implies that we're committing to the path of OS integration over standalone apps ... which I'm all in favor of, and in any case matches industry trends for cloud file sharing. This sounds fantastic on the surface, but maybe we want to have a dedicated conversation about this as whether it is, indeed, our strategic direction.

lidel:
Quick feedback: keep PRs small. Easier to discuss, easier to review, less mistakes :+1: 


- **limit the scope of this PR to the Settings cleanup**   
  - focus on decoupling ipfs-desktop's Settings from ipfs-webui screen and moving them to intuitive menu items
  - keep bundling ipfs-webui and keep opening it in Electron for now
   (switching to browser should be separate PR)
- menu actions that are ON/OFF should indicate clear action (start with verb?):
  - examples: "Enable npm-on-ipfs", "Enable command line tools" etc

hacdias:
@lidel that's actually what I thought after doing this haha. I'm going to do another PR with the preferences section only then!

hacdias:
I'll close this PR.

PR #1419, opened 2020-04-17
feat(ux): improve download cid

### LINKED ISSUE #1135: UI: Not apparent that "Download Hash" uses clipboard
I am glad to see a native UI within the Finder and was curious as to the purpose of the "Download Hash" entry. When I select it, I am told the hash in my clipboard is not valid. I was not aware that the clipboard was going to be used when selecting this menu item. I anticipated some UI to appear allowing me to somehow specify the hash as input.

It would be helpful if the underlying function of this menu option were more apparent without having to select the option and being told of invalid input without knowing the source of the input. Something either indicative that the clipboard will be used or providing some other UI for the user to specify a source of the hash.

ipfs-desktop 0.9.2
macOS 10.14.6

hacdias:
Closes #1135.

I'm working on improving the UX around downloading a CID content. Since Electron does not support prompts natively, I need to implement them! I need to have two color schemes: for dark and light themes that match minimally macOS and Windows so they look relatively good when compared to the native dialogs.

Or we can take a IPFS-looking approach.

Here is what I have for the dark mode now:

image

@jessicaschilling @lidel could you help me with the color schemes? And the copy, please? 😄


assets/locales/en.json
```diff
@@ -41,7 +35,7 @@
   "restartIpfsDesktop": "Restart IPFS Desktop",
   "openLogs": "Open logs",
   "takeScreenshot": "Take Screenshot",
-  "downloadHash": "Download Hash",
+  "downloadHash": "Download CID",
```
jessicaschilling, line 38:
```suggestion
  "downloadHash": "Download from CID...",
```
I think we need the "from" so it's explicit that the user isn't downloading a CID. The ellipses are there to indicate that something will pop up in the next step.

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
   "runGarbageCollectorDone": {
     "title": "Garbage collector",
     "message": "The garbage collector ran successfully."
+  },
+  "downloadCidContentDialog": {
+    "title": "Download",
```
jessicaschilling, line 140:
```suggestion
    "title": "Download from CID",
```

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
...
+    "message": "Type down a CID, IPFS or IPNS path whose contents you want to download.",
+    "action": "Download"
+  },
+  "cantResolveCidDialog": {
+    "title": "Could not resolve",
```
jessicaschilling, line 145: I might give the title of all these error dialogs just "Error" to avoid duplication of content.

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
   "runGarbageCollectorDone": {
     "title": "Garbage collector",
     "message": "The garbage collector ran successfully."
+  },
+  "downloadCidContentDialog": {
+    "title": "Download",
+    "message": "Type down a CID, IPFS or IPNS path whose contents you want to download.",
```
jessicaschilling, line 141:
```suggestion
    "message": "Enter a CID, IPFS path, or IPNS path to download its contents:",
```

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
...
+    "title": "Could not resolve",
+    "message": "The given path ({ path }) could not be resolved."
+  },
+  "couldNotGetCidDialog": {
+    "title": "Could not fetch path",
+    "message": "The given path ({ path }) could not be fetched."
+  },
+  "contentsSavedDialog": {
+    "title": "Contents saved",
+    "message": "The contents from the given path ({ path }) were successfully saved.",
```
jessicaschilling, line 154:
```suggestion
    "message": "The contents of \"{ path }\" were successfully downloaded.",
```

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
...
     "message": "The garbage collector ran successfully."
+  },
+  "downloadCidContentDialog": {
+    "title": "Download",
+    "message": "Type down a CID, IPFS or IPNS path whose contents you want to download.",
+    "action": "Download"
+  },
+  "cantResolveCidDialog": {
+    "title": "Could not resolve",
+    "message": "The given path ({ path }) could not be resolved."
```
jessicaschilling, line 146:
```suggestion
    "message": "Unable to resolve \"{ path }\"."
```
Probably want to avoid the use of the word "path" here in case someone pastes in a CID.

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
...
+  },
+  "couldNotGetCidDialog": {
+    "title": "Could not fetch path",
+    "message": "The given path ({ path }) could not be fetched."
```
jessicaschilling, line 150:
```suggestion
    "message": "Unable to fetch \"{ path }\"."
```

assets/locales/en.json
```diff
@@ -141,5 +135,27 @@
...
+  },
+  "contentsSavedDialog": {
+    "title": "Contents saved",
```
jessicaschilling, line 153:
```suggestion
    "title": "Success",
```

jessicaschilling:
I left some comments on the text -- thank you!
As for colors ... rather than not match *any* OS's dark or light mode, maybe we specifically match one OS color scheme? What's our most frequently used OS?

hacdias:
@jessicaschilling the [downloads](https://github.com/ipfs-shipyard/ipfs-desktop#install) say Windows. The statistics say Linux, which I __doubt__. Even though there are much more sessions on macOS and Windows, it says the number of unique visitors is really small. 

I don't believe 39 visitors can make 31 **thousand** sessions. Something must be wrong on the statistics side. Will open an issue about that.

image

The sessions I believe they're correct, just not the visitors. Wdyt @lidel?

assets/locales/en.json
hacdias: I'm just afraid this would get very long. However, let's also note we can use IPNS or IPFS paths and not only a raw CID. I think we need the most generic naming here, but I'm not sure what it might be.

hacdias: What about "just" "Download..."?

hacdias:
So, after a bit of experimentation. Here's how they look on macOS:

Screenshot 2020-04-18 at 10 39 39

Screenshot 2020-04-18 at 10 39 32

On Windows (the colors actually match better than I thought):

image

image

Unfortunately, Windows and Electron still have some problems recognizing dark themes so there are things such as title bars that are not yet dark, but... we can't do anything about that!

assets/locales/en.json
jessicaschilling: I think this’ll work for now. Worst case is the user gets to the dialog, decides “that’s not what I meant” and cancels.

hacdias: I'll change then. If we end up finding a better message, we can change it anytime 😃

assets/locales/en.json
hacdias: As per the comment above, I'll be keeping this Download.

hacdias:
> label above the input sets expectation about what will happen next?

I don't mind changing, but I think @jessicaschilling should give their opinion here 😃 

> "Download" should be renamed to "Proceed"

Hmm... but we are, effectively, downloading something, aren't we?

> Perhaps it will look better if we reuse ipfs-css in this dialog, make it feel like a part of webui family?

I didn't want to bring ipfs-css and/or tachyons because it would just slow down the window opening. I thought about it, yes, but for that reason, avoided it. We could, certainly, reuse the CSS.

Also, by bringing more external files, we need to take in account how to load them. Right now I'm just feeding a base64 URL with the entire HTML for simplicity.

hacdias:
Just tweaked the colors (used Apple's Blue) and changed the default button color to white.

Screenshot 2020-04-21 at 08 43 43

Screenshot 2020-04-21 at 08 43 53

hacdias:
Colors tweaked, match macOS best (got colors from official Apple Guidelines):

Screenshot 2020-04-21 at 09 22 10

Screenshot 2020-04-21 at 09 22 17

jessicaschilling:
Sorry to miss the thread earlier on the download dialog only being a middle step. I’d suggest that changing the active button to “Next” and removing the phrase “to download its contents” takes care of the problem without having to explicitly explain what’s about to happen.

hacdias:
Text updated:

image

hacdias:
Ping @lidel :)

assets/locales/en.json
```diff
@@ -155,5 +149,27 @@
   "runGarbageCollectorErrored": {
     "title": "Garbage collector",
     "message": "The garbage collector run could not be completed successfully."
+  },
+  "downloadCidContentDialog": {
+    "title": "Download",
```
lidel, line 154: (Just an idea, feel free to ignore): 
what if we set expectations about what "Next" does via the dialog title?

```suggestion
    "title": "Download to a local directory",
```

src/download-cid.js
```diff
@@ -0,0 +1,144 @@
...
+  await fs.outputFile(location, file.content)
+}
+
+async function getCID () {
+  const text = clipboard.readText().trim()
+
+  const { button, input } = await showPrompt({
+    title: i18n.t('downloadCidContentDialog.title'),
+    message: i18n.t('downloadCidContentDialog.message'),
+    defaultValue: isIPFS.cid(text) ? text : '',
```
lidel, line 31: Automatically filling the field if CID is in clipboard is a REALLY nice experience!

`downloadCidContentDialog.message` says this could be a CID or a content path, so perhaps change this to:
```suggestion
    defaultValue: (isIPFS.cid(text) || isIPFS.path(text)) ? text : '',
```

src/download-cid.js
```diff
@@ -0,0 +1,144 @@
...
+    buttons: [
+      i18n.t('downloadCidContentDialog.action'),
+      i18n.t('cancel')
+    ],
+    window: {
+      width: 460,
```
lidel, line 37: Should we make it a tiny bit wider so the entire base32 cid fits? 
On my Linux box it gets cut off, but may be to custom font settings I have:

> 2020-04-22-002352_460x120_scrot

(if it already fits on Mac and Win and is only a Linux issue, ignore this comment)

src/prompt/template.js
```diff
@@ -0,0 +1,95 @@
...
+      button.addEventListener('click', event => {
+        ipcRenderer.send('${id}', {
+          input: document.querySelector('input').value,
+          button: Number(button.id)
+        })
+      })
+    }
+
+    document.querySelector('input').addEventListener('keypress', (event) => {
+      if (event.keyCode == 13) {
```
lidel, line 89: I think `keyCode` was being deprecated for a while:
 https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

@hacdias  please check if we can use `.code` in Electron and switch to that if possible.

src/download-cid.js
jessicaschilling: +1 😊

assets/locales/en.json
jessicaschilling: +1 😊

jessicaschilling:
Not relevant to this PR in particular, but ...

> UX when downloading bigger directories will be awkward (you see no progress and then after a few minutes/hours get a popup telling you download is done 🙃 ), but that can be addressed in the future.

There are a number of issues in our backlog for Desktop/WebUI for various progress feedback indicators. Maybe we want to think about tackling these together for the sake of consistent experience? Not sure where this would fit into our overall prioritization (I'd say P1?) but it might be an interesting problem for @rafaelramalho19 to pull apart?

src/download-cid.js
hacdias: Done!

src/prompt/template.js
hacdias: Updated!

hacdias:
@jessicaschilling @lidel about the UX in this case: with the new ctl, we're getting the new API (#1411), we'll be able to get "some" progress when downloading directories. Maybe we can do something like what we do with the GC and update the status with some informational message.

What we could do too was to update the icon on the tray and show some loading animation, but that requires more work and needs to be ensured to work on Windows, macOS and the most used linux distros.

And yes, @jessicaschilling, I think it would be nice to have one issue of that kind on each repository (webui and desktop), aggregating the feedback UX improvements on each app.

---

Just a little FYI: even though I've been quite dedicated to Desktop/Web UI this past week, I'm not yet _back_ to PL from my break to finish my bachelor's degree.  I'll always try to reply to everything since this covid-situation is making me have a bit more time and that's why I've been dedicating a few hours on fixing issues here 😄

jessicaschilling:
@hacdias any time you need more joy in your life, you're more than welcome to come hang out in gui-land, haha!

PR #1420, opened 2020-04-18
feat(ux): improve gc

hacdias:
As @RubenKelevra [noted](https://github.com/ipfs-shipyard/ipfs-desktop/pull/1407#issuecomment-615677476), GC might take a while to finish running. So I took their suggestions and implemented them (see bellow).

This PR:
- Improves UX while GC is running;
- Adds a custom error message when GC fails;
- Cleans up tray logic a bit.

License: MIT


RubenKelevra:
It's neat for blocking this option, but we should show it more prominently.

How about changing the green dot to a yellow broom, and instead of "IPFS is running" change the message to "IPFS is running the garbage collector" or something like this? 🤔

I think the ususally degraded performance justifies that the icon should be yellow instead of green.

hacdias:
@RubenKelevra that's definitely a good idea! Let me see what I can do. I actually wanted to clean up tray logic. Might be an opportunity to improve it a bit.

hacdias:
GC not running:

Screenshot 2020-04-18 at 09 29 15

GC running:

Screenshot 2020-04-18 at 09 30 02

RubenKelevra:
Well, I would still vote for a broom instead of just 'yellow' which looks like something is partly broken 🤔

But the rest looks very nice!

hacdias:
Ahah I like the Broom idea but we'd need to add that icon at some specific sizes and I think it's easier for now to keep the yellow circle. 🧹🧹

RubenKelevra:
> Ahah I like the Broom idea but we'd need to add that icon at some specific sizes and I think it's easier for now to keep the yellow circle.

Here you go:

https://commons.wikimedia.org/wiki/File:Artworkbean_broom.svg

hacdias:
Thanks! Let's see what @lidel and @jessicaschilling have to say

jessicaschilling:
This is cool! I’m a little concerned that users will think IPFS isn’t running when GC is running. There isn’t a great option for text due to space but maybe “Running with Garbage Collection”? And I’d gray out the stop and restart items rather than remove them entirely, to indicate you can still stop and restart, just not right now. 

An aside: We should capitalize “IPFS is **R**unning” to match the other menu items. 

Regarding broom - it’s going to be really tough to make that legible at small sizes, unfortunately. We’re also working on some stuff in ipfs-css that’s nearly finished, so suggest leaving that out for the time being.

hacdias:
> I’m a little concerned that users will think IPFS isn’t running when GC is running. There isn’t a great option for text due to space but maybe “Running with Garbage Collection”? 

What if we explain in the dialog before? I'm concerned about such a long sentence because it would seem all the other options are really small compared to that one.

> An aside: We should capitalize “IPFS is Running” to match the other menu items.

Will do!
 
> Regarding broom - it’s going to be really tough to make that legible at small sizes, unfortunately. 

👍

jessicaschilling:
“Running (Garbage Collecting)”?
If it takes a long time, they might forget about the dialog 😉

hacdias:
image

Like this?

jessicaschilling:
I see what you mean, but I do think it’s better than being unclear. Plus it’s not a super-frequent scenario. Let’s see what @lidel says.

hacdias:
Yeah, I agree. Let's wait for lidel input!

assets/locales/en.json
```diff
@@ -1,9 +1,10 @@
...
-  "ipfsIsStopping": "IPFS is stopping",
-  "ipfsIsStarting": "IPFS is starting",
-  "ipfsIsNotRunning": "IPFS is not running",
-  "ipfsHasErrored": "IPFS has errored",
+  "ipfsIsRunning": "IPFS is Running",
+  "ipfsIsStopping": "IPFS is Stopping",
+  "ipfsIsStarting": "IPFS is Starting",
+  "ipfsIsNotRunning": "IPFS is Not Running",
+  "ipfsHasErrored": "IPFS has Errored",
+  "runningWithGC": "Running (Garbage Collecting)",
```
lidel, line 7: > (drive by comment, will do proper review on Monday)

Not sure why, but below reads more "natural" to me:

```suggestion
  "runningWithGC": "Running (GC in progress…)",
```

Thoughts @jessicaschilling?

hacdias: The biggest concern might be if the user knows or not what 'GC' means. However, if they're running the GC, it's highly likely that they already knew what it was and does.

jessicaschilling: Agreed that it seems more natural. And I think what @hacdias brings up about putting "(GC)" in the preceding dialog will help considerably.

The only thing about this plan I would change is remove the "..." in the "Running" item, because in the context of menu items that implies "choose this menu item and then we give you a dialog with more info".

hacdias: Just updated it! 

image

image

jessicaschilling: I'm down if @lidel is. 😊

assets/locales/en.json
```diff
@@ -149,5 +150,9 @@
   "cantAddIpfsToPath": {
     "title": "IPFS on PATH",
     "message": "Could not add IPFS to the PATH."
+  },
+  "runGarbageCollectorErrored": {
+    "title": "Garbage collector",
+    "message": "The garbage collector ran could not be completed successfully."
```
lidel, line 156: A typo?
```suggestion
    "message": "The garbage collector run could not be completed successfully."
```

RubenKelevra:
> Regarding broom - it’s going to be really tough to make that legible at small sizes, unfortunately.
Maybe doing a small animation would help?

Here's some inspiration :)

source

PR #1421, opened 2020-04-18
feat(ux): better error messages

hacdias:
This adds the possibility of adding custom error messages. I also updated some places to show specific error messages instead of general ones.

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -141,5 +141,13 @@
   "runGarbageCollectorDone": {
     "title": "Garbage collector",
     "message": "The garbage collector ran successfully."
+  },
+  "moveRepositoryFailed": {
+    "title": "Move repository",
+    "message": "Could not move the repository from { currDir } to { newDir }."
+  },
+  "cantAddIpfsToPath": {
+    "title": "IPFS on PATH",
+    "message": "Could not add IPFS to the PATH."
   }
 }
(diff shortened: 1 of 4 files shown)


assets/locales/en.json
```diff
@@ -141,5 +141,13 @@
   "runGarbageCollectorDone": {
     "title": "Garbage collector",
     "message": "The garbage collector ran successfully."
+  },
+  "moveRepositoryFailed": {
+    "title": "Move repository",
+    "message": "Could not move the repository from { currDir } to { newDir }."
```
jessicaschilling, line 147: Put double quotes around the values of currDir and newDir? (Sorry to not suggest code itself, am on my phone!)

assets/locales/en.json
```diff
@@ -141,5 +141,13 @@
...
+  },
+  "cantAddIpfsToPath": {
+    "title": "IPFS on PATH",
```
jessicaschilling, line 150: “IPFS on path”? Not sure why we’d uppercase “path” here.

assets/locales/en.json
```diff
@@ -141,5 +141,13 @@
...
+    "message": "Could not add IPFS to the PATH."
```
jessicaschilling, line 151: “... to the path”? Not sure why we’d uppercase “path” here.

assets/locales/en.json
hacdias: We're referring to the environment variable PATH so we usually put that in uppercase 😃

jessicaschilling: Ah right, never mind!

PR #1425, opened 2020-04-18
feat(ux): move preferences to menubar

### LINKED ISSUE #1436: Disable WebUI links in "IPFS is Not Running" state
When node is stopped/offline, menu items that open WebUI are still active:

> off-when-off-2020-04-23--00-49-45

Opening them shows webui in "unable to connect to node" state, which is bit confusing for less technical users.

I believe we should make them inactive when "IPFS is Not Running"

hacdias:
This PR removes the Desktop settings from the Web UI panel! Related to #1389.

- See https://github.com/ipfs-shipyard/ipfs-webui/pull/1445 **merged**
- macOS guidelines prefer the term 'Preferences', others OSes users prefer 'Settings'
- Adds the settings under 'Preferences' (macOS) or 'Settings' (others)
- Some that might require more explanation get an information dialog before installing/enabling.
- Errors are now more visible (user gets a dialog). Before, Web UI only said: "error while enabling X".
- Closes #1436 (disables web ui links when node is offline)

Screenshot 2020-04-18 at 18 45 44

License: MIT

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -17,7 +17,7 @@
   "readReleaseNotes": "Read Release Notes",
   "status": "Status",
   "files": "Files",
-  "settings": "Settings",
+  "peers": "Peers",
   "quit": "Quit",
   "versions": "Versions",
   "screenshotTaken": "Screenshot taken",
@@ -32,6 +32,7 @@
   "close": "Close",
   "ok": "OK",
   "cancel": "Cancel",
+  "enable": "Enable",
   "reportTheError": "Report the error",
   "restartIpfsDesktop": "Restart IPFS Desktop",
   "openLogs": "Open logs",
@@ -171,5 +172,56 @@
   "couldNotSaveDialog": {
     "title": "Could not write to disk",
     "message": "There was an error writing to the disk. Please try again."
+  },
+  "launchAtLoginNotSupported": {
+    "title": "Error",
+    "message": "Launch at login is not supported on your platform."
+  },
+  "launchAtLoginFailed": {
+    "title": "Error",
+    "message": "Launch at login could not be enabled on your machine."
+  },
+  "enableIpfsOnPath": {
+    "title": "Enable IPFS on PATH",
+    "message": "By enabling this option, IPFS will be available on your command line as \"ipfs\". This action is reversible.",
+    "action": "Enable"
+  },
+  "disableIpfsOnPath": {
+    "title": "Disable IPFS on PATH",
+    "message": "By disabling this option, IPFS will no longer be available on your command line as \"ipfs\".",
+    "action": "Disable"
+  },
+  "enableGlobalTakeScreenshotShortcut": {
+    "title": "Enable screenshot shortcut",
+    "message": "By enabling this, the shortcut \"{ accelerator }\" will be available to take screenshots as long as IPFS Desktop is running."
+  },
+  "enableGlobalDownloadShortcut": {
+    "title": "Enable download shortcut",
+    "message": "By enabling this, the shortcut \"{ accelerator }\" will be available to download files as long as IPFS Desktop is running."
+  },
+  "installNpmOnIpfsWarning": {
+    "title": "Install npm on IPFS",
+    "message": "This experimental feature installs the \"ipfs-npm\" package on your system. It requires Node.js to be installed.",
+    "action": "Install"
+  },
+  "unableToInstallNpmOnIpfs": {
+    "title": "Error",
+    "message": "It was not possible to install \"ipfs-npm\" package on your system. Please check the logs for more information or try installing it manually by running \"npm install -g ipfs-npm\" on your command line."
+  },
+  "unableToUninstallNpmOnIpfs": {
+    "title": "Error",
+    "message": "It was not possible to uninstall \"ipfs-npm\" package on your system. Please check the logs for more information or try uninstalling it manually by running \"npm uninstall -g ipfs-npm\" on your command line."
+  },
+  "settings": {
+    "settings": "Settings",
+    "preferences": "Preferences",
+    "openNodeSettings": "Open Node Settings",
+    "desktopIntegrations": "Desktop Integrations",
+    "launchOnStartup": "Launch at Login",
+    "ipfsCommandLineTools": "Command Line Tools",
+    "takeScreenshotShortcut": "Global Screenshot Shortcut",
+    "downloadHashShortcut": "Global Download Shortcut",
+    "experiments": "Experiments",
+    "npmOnIpfs": "npm on IPFS"
   }
 }
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -216,7 +216,7 @@
     "settings": "Settings",
     "preferences": "Preferences",
     "openNodeSettings": "Open Node Settings",
-    "desktopIntegrations": "Desktop Integrations",
+    "appPreferences": "App Preferences",
     "launchOnStartup": "Launch at Login",
     "ipfsCommandLineTools": "Command Line Tools",
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "clean:webui": "shx rm -rf assets/webui/",
     "build": "run-s clean:webui build:*",
     "build:webui": "run-s build:webui:*",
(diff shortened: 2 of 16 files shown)


assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
   "cantAddIpfsToPath": {
     "title": "IPFS on PATH",
     "message": "Could not add IPFS to the PATH."
+  },
+  "launchAtLoginNotSupported": {
+    "title": "Launch at login",
```
jessicaschilling, line 155:
```suggestion
    "title": "Error",
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+    "message": "Launch at login is not supported on your platform."
+  },
+  "launchAtLoginFailed": {
+    "title": "Launch at login",
```
jessicaschilling, line 159:
```suggestion
    "title": "Error",
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+    "message": "Launch at login could not be enabled on your machine."
+  },
+  "ipfsOnPathInstall": {
+    "title": "Install IPFS on PATH",
+    "message": "By enabling this option, IPFS will be available on your command line as \"ipfs\". This action is reversible.",
+    "action": "Install"
```
jessicaschilling, line 165: For both the dialog boxes (and their menu items) related to IPFS on PATH, we're using "install/enable" and "uninstall/disable" interchangeably, which is a little confusing. Let's decide on either install/uninstall or enable/disable and be consistent throughout ... I might go for enable/disable so that it feels more like an option rather than something more severe, but there might be precedence for install/uninstall elsewhere that I'm unaware of?

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+  },
+  "ipfsOnPathUninstall": {
+    "title": "Uninstall IPFS on PATH",
+    "message": "By disabling this option, IPFS will no longer be available on your command line as \"ipfs\".",
+    "action": "Uninstall"
+  },
+  "enableGlobalTakeScreenshotShortcut": {
+    "title": "Take screenshot shortcut",
```
jessicaschilling, line 173:
```suggestion
    "title": "Enable screenshot shortcut",
```
(assuming we are using "enable" language)

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+    "message": "By enabling this, the shortcut { accelerator } will be available to take screenshots as long as IPFS Desktop is running."
```
jessicaschilling, line 174:
```suggestion
    "message": "By enabling this, the shortcut \"{ accelerator }\" will be available to take screenshots as long as IPFS Desktop is running."
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+  },
+  "enableGlobalDownloadShortcut": {
+    "title": "Download shortcut",
```
jessicaschilling, line 177:
```suggestion
    "title": "Enable download shortcut",
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+    "message": "By enabling this, the shortcut { accelerator } will be available to download files as long as IPFS Desktop is running."
```
jessicaschilling, line 178:
```suggestion
    "message": "By enabling this, the shortcut \"{ accelerator }\" will be available to download files as long as IPFS Desktop is running."
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+  },
+  "settings": {
+    "settings": "Settings",
+    "preferences": "Preferences",
+    "launchOnStartup": "Enable Launch at Login",
+    "ipfsCommandLineTools": "Enable Command Line Tools",
+    "takeScreenshotShortcut": "Enable Take Screenshot Shortcut",
```
jessicaschilling, line 185:
```suggestion
    "takeScreenshotShortcut": "Enable Screenshot Shortcut",
```

assets/locales/en.json
```diff
@@ -149,5 +150,41 @@
...
+    "downloadHashShortcut": "Enable Download Hash Shortcut",
```
jessicaschilling, line 186:
```suggestion
    "downloadHashShortcut": "Enable Download Shortcut",
```

hacdias:
@jessicaschilling thanks for pointing that out. They’re actually not enabled. Those options enable them globally. They are checkboxes.

Now that I think of it, being a checkbox and having a verb (“Enable...”/“Disable...”) does not seem to be a great idea. 

Either we keep the checkbox and remove the verb (simpler solution) or we toggle between verbs (Enable and Disable).

jessicaschilling:
Checkboxes are good! If unchecked, do “take screenshot” and “disable hash” menu items act accordingly (do the shortcuts disappear)?

hacdias:
> Checkboxes are good! If unchecked, do “take screenshot” and “disable hash” menu items act accordingly (do the shortcuts disappear)?

The shortcuts **always** work if the menu is open. That’s why they’re visible. The preferences enable them globally, ie, even with the menu closer, you can press them and they’ll work. I guess we should find a better way to pass that message.

jessicaschilling:
OH ... got it. Sorry! Good point though; I was confused, others will be too.

We don’t actually need “Enable” in menu items - the checkboxes imply it (see your Lungo example). Removing it would allow room for “Global Screenshot Shortcut” and “Global Download Shortcut”, which might be more explicit. Would like to run that past @lidel though.

hacdias:
Yes, I agree we should wait for @lidel opinion on this 😃

hacdias:
image

assets/locales/en.json
hacdias: Just updated everywhere to enable/disable that is not related to IPFS Desktop updates 😃 

>  I might go for enable/disable so that it feels more like an option rather than something more severe

Agreed. We're not actually installing anything but when updating or enabling `npm on IPFS`, which I will make more clear.

hacdias:
Just as a thought: I wonder if some of the 'Advanced' things shouldn't be moved under Preferences. The only one that doesn't make much sense to me would be 'Run Garbage Collector'. All the others seem to fit under Preferences. Wdyt? (We can leave this to another PR if this one is good)

image

jessicaschilling:
@hacdias - I'd be inclined to leave these under "Advanced" ... to a certain degree, they're sort of applicable to being considered preferences ... but only to advanced users. ;)

assets/locales/en.json
```diff
@@ -16,7 +16,7 @@
   "readReleaseNotes": "Read Release Notes",
   "status": "Status",
   "files": "Files",
-  "settings": "Settings",
```
lidel, line 19: Removing this from the top level menu may be a problem.  I believe users will look for the "WebUI Settings" screen under "Settings" and won't find any entry point there. 

Can we add it at the very top, to maintain continuity, of sorts? 

One entry followed by a separator and then existing entries under "Desktop" header (similar to existing  "Experiments"):

> 


@jessicaschilling  Does this make sense? If so, what would be the correct wording here: "Open Node Settings"? "Open WebUI Settings"?

hacdias: Just rebased on top of master and added a "Open Node Settings" which opens Web UI settings:

image

jessicaschilling: This is great. Agree on “Node Settings” over “WebUI settings sine the thing you’re adjusting is the node itself.

lidel: What about adding `Desktop integrations` header below the separator, to make it even more clear?

> 
png

hacdias: Just added:
image


Just wondering if it's really needed. But looks great

jessicaschilling: Having a clarification of “these are the prefs for this app” is a good idea. I might just say “App Preferences” though? Not sure if “desktop integrations” makes sense unless you’re already deeply familiar.

hacdias: @lidel what do you think?

hacdias:
@lidel other than the the 'Desktop Integrations' label, is there anything you'd change?

hacdias:
Just renamed 'Desktop Integrations' to 'App Preferences'

image

jessicaschilling:
LGTM, but let's get @lidel's blessing first 😊

hacdias:
Yes!

hacdias:
@lidel I decided to fix #1436 because those links enable the user to access IPFS Desktop settings unless this one gets merged.

PR #1426, opened 2020-04-18
docs: clarify protocol usage

### LINKED ISSUE #1366: Update readme to clarify how links can be opened
The readme mentions

"Handle ipfs://, ipns:// and dweb: links

If you come across a link to any of the protocols above, IPFS Desktop will be able to open them and redirect them to your default browser."

but it is missing, both from here as from ipfs guides (https://docs.ipfs.io/guides), how should one proceed to open an ipfs link that is provided as text. This is a very basic feature and I fear that if I'm having this much trouble, this could scare a lot of users from further digging into ipfs. If I have an ipfs link, say, /ipns/torrent-paradise.ml, how should I proceed to open it using the ipfs-desktop application or the background daemon?

Edit: with the frontend, I'm of course able to visualize the files from a web application, but how can I open the link directly in the browser as a website? I know I could do it through cloudfare or https://ipfs.io/LINK, but these are just centralized proxy services to the actual descentralized network, so it is important to be able to open the links in the browser with some 127.0.0.1/ipfs/obj syntax. This is missing from documentation though.

Edit2: I kept randomly messing with the port numbers until I found the correct one: 8080. So, in the example given, `http://127.0.0.1:8080/ipns/<something>` typed into the browser will show the website. Although I figured it out by sheer luck, I still think it is very important to have this info in the readme. 8080 is a common port providing a "localhost" alternative for the 80 http default, but in no way should be taken as a given.

hacdias:
Closes #1366.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ IPFS Desktop's main feature is to allow you to have the IPFS daemon always runni
 
 ### Handle `ipfs://`, `ipns://` and `dweb:` links
 
-IPFS Desktop has the capacity to handle protocols at the Operating System level, i.e., if an application tries to open a link to either `ipfs://`, `ipns://` or `dweb:`, then the OS will call IPFS Desktop, which will handle the link.
+IPFS Desktop has the capacity to handle protocols at the operating system level; i.e., if an application tries to open a link to either `ipfs://`, `ipns://` or `dweb:`, then the OS will call IPFS Desktop, which will handle the link.
 
 For example, if you come across a link on the Internet whose `href` attribute is `ipns://ipfs.io`, then IPFS Desktop will be able to handle it.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,11 @@ IPFS Desktop's main feature is to allow you to have the IPFS daemon always runni
 
 ### Handle `ipfs://`, `ipns://` and `dweb:` links
 
-If you come across a link to any of the protocols above, IPFS Desktop will be able to open them and redirect them to your default browser.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,7 +40,11 @@ IPFS Desktop's main feature is to allow you to have the IPFS daemon always runni
 
 ### Handle `ipfs://`, `ipns://` and `dweb:` links
 
-If you come across a link to any of the protocols above, IPFS Desktop will be able to open them and redirect them to your default browser.
+IPFS Desktop has the capacity to handle protocols at the Operating System level, i.e., if an application tries to open a link to either `ipfs://`, `ipns://` or `dweb:`, then the OS will call IPFS Desktop, which will handle the link.
```
jessicaschilling, line 43:
```suggestion
IPFS Desktop has the capacity to handle protocols at the operating system level; i.e., if an application tries to open a link to either `ipfs://`, `ipns://` or `dweb:`, then the OS will call IPFS Desktop, which will handle the link.
```

README.md
```diff
@@ -40,7 +40,11 @@ IPFS Desktop's main feature is to allow you to have the IPFS daemon always runni
...
+
+For example, if you come across a link on the Internet whose `href` attribute is `ipns://ipfs.io`, then IPFS Desktop will be able to handle it.
+
+On Windows, an easy way to open a link would be to open the Run window (Win+R), paste the link and hit enter. Then, IPFS Desktop handles the rest. The same happens if you try opening the same link in a browser.
```
jessicaschilling, line 47:
```suggestion
On Windows, an easy way to open a link would be to open the Run window (Win+R), paste the link, and hit Enter. Then, IPFS Desktop handles the rest. The same happens if you try opening the same link in a browser.
```

PR #1427, opened 2020-04-18
feat: allow custom ipfs binary

### LINKED ISSUE #836: Maybe provide a way to configure which binary to use?
[Advanced users](https://ipfs.io/ipfs/QmY2EtB8HPKkL87UyBR1tchrm8FhBQbrYrkNg3nPg56cBZ/1be2s7.jpg) will want to run newer and newer versions of `ipfs` (eg release candidates, etc). 

While i'm perfectly comfortable replacing the binary at `/Applications/IPFS Desktop.app/Contents/Resources/app.asar.unpacked/node_modules/go-ipfs-dep/go-ipfs/ipfs` with a shell script that uses the `ipfs` in my `$PATH`, it may be nicer to make it a gui config option.

Though, maybe warn users that it's an advanced feature and that things may start breaking for them (for example, if the installed version doesn't work with the api client IPFS Desktop uses, or the binary in the path is broken, etc), which may be hard to recover from.

Just a suggestion-- perhaps the best answer is "no, do your shell script replace manually, and deal with the consequences" to avoid bad things.

hacdias:
This closes #836. With this PR, the user is now able to set their own custom IPFS binary. The option to do so is under Advanced as it is an advanced feature:

image

If a custom IPFS binary is set, then the 'Set Custom IPFS Binary' is replaced by 'Clear Custom IPFS Binary'.

The IPFS on `PATH` now obbeys to this so the `ipfs` binary available for the user is the one that IPFS Desktop is using, no matter if it's provided by `IPFS_GO_EXEC`, set by the user, or the default that is bundled with IPFS Desktop.

Setting the environment variable `IPFS_GO_EXEC` affects the IPFS binary used while it is set. If IPFS Desktop is restarted and that variable is no longer set, we default to either the custom user defined binary or to the bundled binary.

License: MIT


hacdias:
> This PR is a good opportunity to unify / double check those behaviors.

I wanted to add an information here about what we are doing about the `IPFS_PATH` environment variable: we are **not** setting a global `IPFS_PATH` variable. We're just overriding it when using the "IPFS on PATH" feature.

> menu items should be next to each other, perhaps separated from the rest with a line?

Just done.

> changes made via GUI should persist across restarts

Already done.

> changes made via CLI disappear when env variable is no longer set during next run

We're not setting global env vars.

> this really needs tests

It does. And they're complicated. Windows, macOS and Linux.

hacdias:
I think I interpreted you wrong @lidel! The env vars should affect IPFS Desktop. And not the opposite. About `IPFS_PATH` then: the variable defines a repository and then we store on the config the repository of `ipfsd.path`. So, if `IPFS_PATH` was chosen initially to be the repository on the first run, that'll be what we save on the config.

hacdias:
@lidel tested and works on macOS and Windows. I'd like to hear some suggestions on how to automatically test this.

PR #1429, opened 2020-04-20
feat(ux): open Web UI at launch on Linux

hacdias:
See https://github.com/ipfs-shipyard/ipfs-desktop/issues/1153#issuecomment-616243802.

Adds the option to open Web UI automatically on startup. This is enabled by default when the OS is neither Windows, nor macOS to help users that do not have a tray icon.

License: MIT

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -219,7 +219,7 @@
     "openNodeSettings": "Open Node Settings",
     "appPreferences": "App Preferences",
     "launchOnStartup": "Launch at Login",
-    "launchWebUIOnStartup": "Launch Web UI on Startup",
+    "openWebUIAtLaunch": "Open Web UI at Launch",
     "ipfsCommandLineTools": "Command Line Tools",
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -219,6 +219,7 @@
     "openNodeSettings": "Open Node Settings",
     "appPreferences": "App Preferences",
     "launchOnStartup": "Launch at Login",
+    "launchWebUIOnStartup": "Launch Web UI on Startup",
     "ipfsCommandLineTools": "Command Line Tools",
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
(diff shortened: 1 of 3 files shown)


assets/locales/en.json
```diff
@@ -218,6 +218,7 @@
     "openNodeSettings": "Open Node Settings",
     "appPreferences": "App Preferences",
     "launchOnStartup": "Launch at Login",
+    "launchWebUIOnStartup": "Launch Web UI on Startup",
```
jessicaschilling, line 221: Is it correct that this is different from “Launch at Login” (above) in that the former just starts Desktop in the background, but this launches Desktop and opens the Electron window as well? And that they’re mutually exclusive? If so, maybe the options should read (note order switches):
- Launch on Startup
- Background Launch on Startup

Note: This would mean re-translating the existing string in Transifex.

hacdias: This option means "open Electron window when IPFS Desktop starts" so they're not mutually exclusive

jessicaschilling: Sorry, I was wrong about "mutually exclusive" ... but is "launch with open Electron window on startup" vs "launch on startup but I don't need the Electron window, just the daemon and the menubar options" what we're asking here? If that's the case, I think the launch/background launch options still apply. Unless I'm just missing something really obvious! 😆

hacdias: I'm sorry, my English isn't the best some times. Let me try again to explain better, hopefully haha

- launchOnStartup --> Launch IPFS Desktop when the user logs in
- launchWebUIOnStartup -> Open the electron window (web ui) when IPFS desktop starts

For example, you can have the second one enabled without the first. Then, when you open IPFS Desktop, the web ui window opens too.

If you have both enabled, Desktop starts when the user logs in and the Web UI window pops up too.

If you have the first enabled and the second disabled, then IPFS Desktop starts when you login but you get no Window.

I hope that's clearer. And sorry again 😅

jessicaschilling: Ah, OK. I think I was missing the distinction between "system startup" and "IPFS Desktop startup". This makes more sense. Maybe then:

- "Launch at Login"
- "Open Web UI at Launch"

Using "launch" to bridge those two things together should clarify. Does that sound good to you and @lidel?

Side note: This is one of those situations where saying "Web UI" doesn't really make sense to anyone other than us. In a context like this one, I'd expect to see "Dashboard" or a similar term ... I know we've got other issues open to discuss this, though. Just pointing out an example 😊

hacdias: I agree with the Launch part and the Web UI part too. I agree we should not use Web UI on this context or on any UI at all. I like the 'Dashboard' idea.

jessicaschilling: I'd like to get @lidel's opinion on whether it makes sense to *start* saying "Dashboard" here when we use the phrase "Web UI" in various places already ... or whether we should have a separate effort to start deprecating the language "Web UI" globally. Both have pros and cons.

hacdias: Just leaving this here as a ref: https://github.com/ipfs-shipyard/ipfs-webui/issues/612

jessicaschilling: Based on that, probably worth leaving as "Web UI" until we have @olizilla back for more in-depth discussion.

hacdias: Just updated to "Open Web UI at Launch"

lidel:
Rebased to be sure it works with the latest master and tests still pass.

PR #1438, opened 2020-04-23
fix: repo fsck is deprecated

hacdias:
`ipfs repo fsck` seems to have been deprecated (https://github.com/ipfs/go-ipfs/commit/288a83ce7dcbf4a2498e06e4a95245bbb5e30f45) in the meanwhile so our logic wasn't working.

I changed it so now we manually check if it's a connection refused error. If so, check if there's a config file. If so, remove the api file.

If no config file, then it's a remote api connection, show error to the user.

License: MIT

### DIFF
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -17,6 +17,10 @@ function apiFileExists (ipfsd) {
   return fs.pathExistsSync(join(ipfsd.path, 'api'))
 }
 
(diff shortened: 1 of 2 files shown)


hacdias:
I'm merging this for now since I'm stumble on this error every now and then and it's annoying. It's also a simple PR. We can change anything on another PR.

Issue #844, opened 2019-02-26
Web UI and Desktop languages


hacdias:
Right now, Desktop gets its language from the operating system its running on and passes it to the Web UI. Although, the user can change the language of the Web UI via its Settings, which does not affect the language of IPFS Desktop and then we get a mixed UI.

I've got some possible options to solve this issue:

| Option | Pros | Cons |
| ------ | ---- | ---- |
|  Change Desktop's language according to Web UI's | Easy to find on settings and simple for the user | We can't make sure if there are the same languages for both the Web UI and Desktop |
| Hide the language option on Web UI and just use the OS language | No problems whatsoever | The user can't pick a different language |
| Create some kind of menu that is smart and knows the available languages for both Web UI and Desktop and only shows the ones that exist for both | Works best | More work, but feasible I think. Harder to keep up to date with all translations. |

@olizilla @fsdiogo @lidel what do you think?

lidel:
I believe the goal is to eventually reach translation parity. 
To get there we _should_ surface missing translations, so potential contributors see what is missing.

I'd go with low hanging fruit: "Change Desktop's language according to Web UI's".
If desktop is missing language selected in Web UI, it should fallback to English strings. 

If we want to go extra mile, Desktop could also display a message asking user if they would like to contribute missing translation, clicking Yes would redirect them to https://www.transifex.com/ipfs/ipfs-desktop/)

hacdias:
First part is done, we now need to drive the extra mile.

> Desktop could also display a message asking user if they would like to contribute missing translation, clicking Yes would redirect them to [transifex.com/ipfs/ipfs-desktop](https://www.transifex.com/ipfs/ipfs-desktop/))

hacdias:
This is something trivial to do. @lidel @jessicaschilling do you think this is still useful?

jessicaschilling:
Yes!

hacdias:
Maybe a dialog when a person opens IPFS Desktop for the first time and then leaving a button on the menu to contribute?

jessicaschilling:
Only for folks for whom their chosen WebUI language doesn't exist in Desktop, right? If so, that sounds good.

hacdias:
Uh, I get it now and it may be more complicated than what I thought then. I'll investigate, but it is certainly possible.

lidel:
I think https://github.com/ipfs-shipyard/ipfs-desktop/pull/1447 is good enough to close this issue. :grimacing:  

[...]

PR #1447, opened 2020-04-25
feat: add "Help Translation" under "About"

hacdias:
This adds "Help Translating" to the menubar "About" section and replaces all "..." by the ellipsis character "…". The link opens [transifex](https://www.transifex.com/ipfs/ipfs-desktop/).

This partially addresses #844. The dialog will come on a different PR.

image


assets/locales/en.json
```diff
@@ -13,7 +13,8 @@
   "openRepoDir": "Open Repository Directory",
   "openLogsDir": "Open Logs Directory",
   "openConfigFile": "Open Configuration File",
-  "viewOnGitHub": "View on GitHub",
+  "viewOnGitHub": "View on GitHub…",
+  "helpUsTranslate": "Help Translating…",
```
jessicaschilling, line 17: “Help Translate This App...”?
I understand it’s going to expand the menu width, but it’s the shortest phrase I can find that makes it clear we are talking about human languages rather than programming languages.

Another option might be “IPFS Translation Project...”, but it doesn’t have a verb in there, which makes it inconsistent with other menu items.

hacdias: Uh, I prefer "Help Translate This App" on this case.

hacdias: I don't think it looks bad:

image

jessicaschilling: Agree! Thanks.

PR #1448, opened 2020-04-26
Updated german locale file

Nerd2201:
Translated some lines from english to german and will continue to do so
within the next days, may I also start to learn spanish...

### DIFF
--- a/assets/locales/de.json
+++ b/assets/locales/de.json
@@ -1,10 +1,10 @@
 {
-  "ipfsIsRunning": "IPFS is Running",
-  "ipfsIsStopping": "IPFS is Stopping",
-  "ipfsIsStarting": "IPFS is Starting",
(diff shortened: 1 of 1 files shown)


jessicaschilling:
Thank you for contributing translation! :+1: 

To avoid duplicated work, we don't accept direct PRs with locales.

Did you submit those translations using Transifex project at https://www.transifex.com/ipfs/ipfs-webui/ (Web UI’s translations also feed Desktop’s)? If not, let us know your Transifex account name and we can add you to the appropriate translation teams so you can contribute there.

FYI, there is no need for submitting PRs via GitHub: updating them on Transifex is enough. We synchronize translations with Transifex before each release. 

You can find more information at [README#translations](https://github.com/ipfs-shipyard/ipfs-webui#translations) and [ipfs/i18n](https://github.com/ipfs/i18n).

PR #1450, opened 2020-04-27
fix(ux): show dock when prompt is visible

hacdias:
Show dock while the prompt is visible.

License: MIT


hacdias:
Just a FYI: I tested and it works. But yes, it's better if someone else checks too :)

rafaelramalho19:
Tested and it works 👍

PR #1452, opened 2020-04-27
fix: keep track of last opened page

### LINKED ISSUE #1451: "Open Web UI at Launch" stays at  `/blank` instead of `/`
WebUI seems to have some leftovers related to settings, which produce an error:

> 2020-04-27-135139_1330x435_scrot

hacdias:
Fixes #1451. Keeps track of the last opened page so every time we need to refresh because a change on the IPFS status, we can keep the user on the same page. The visibility (`/blank` handling) is taken care on the `preload.js`.

License: MIT

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -96,12 +96,17 @@ module.exports = async function (ctx) {
 
   ctx.webui = window
 
-  ctx.launchWebUI = (url, { focus = true } = {}) => {
-    if (!url) {
+  const url = new URL('/', 'webui://-')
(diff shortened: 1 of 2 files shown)


src/webui/index.js
```diff
@@ -126,6 +127,7 @@ module.exports = async function (ctx) {
       apiAddress = ipfsd.apiAddr
       url.searchParams.set('api', apiAddress)
       updateLanguage()
+      console.log(url.toString())
```
lidel, line 130:
```suggestion
```

PR #1476, opened 2020-04-30
Update zh-CN.json

CsterKuroi:


lidel:
Thank you for contributing translation @CsterKuroi! :+1: 

To avoid duplicated work, we don't accept direct PRs with locales.

Did you submit those translations using Transifex projects at https://www.transifex.com/ipfs/ipfs-desktop/  and https://www.transifex.com/ipfs/ipfs-webui/ ? If not, let us know your Transifex account name and we can add you to the appropriate translation teams so you can contribute there.

FYI, there is no need for submitting PRs via GitHub: updating them on Transifex is enough. We synchronize translations with Transifex before each release. 

You can find more information at [README#translations](https://github.com/ipfs-shipyard/ipfs-webui#translations) and [ipfs/i18n](https://github.com/ipfs/i18n).

PR #1478, opened 2020-04-30
docs: improve error reporting template

lidel:
This PR improves Bug reporting template by adding instruction on where to find error logs and asking reporter to attach them to the bug report.

This should save us the usual 1-2 roundtrips in most of  cases.

Thank you to @achingbrain for inspiration :-)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,6 +6,8 @@ labels: need/triage
 assignees: ''
 
 ---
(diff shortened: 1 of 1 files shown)


jessicaschilling:
Great idea, thank you?
Should we ask them their OS and version if they can’t provide logs for some reason? I don’t want to make things too complex, though.

lidel:
I think its better to ask for versions upfront, saves roundtrip.
If they are not savvy enough, they will just leave it blank :)

PR #1501, opened 2020-05-13
chore: update dependencies

### LINKED ISSUE #1500: Switch to go-ipfs 0.5.1
We should release ipfs-desktop v0.11.3 with go-ipfs v0.5.1
Ref. https://github.com/ipfs/go-ipfs/releases/tag/v0.5.1

Note: this depends on https://www.npmjs.com/package/go-ipfs-dep being updated to 0.5.1 first, which should happen automatically within 1-2h after https://dist.ipfs.io is updated. If its not at 0.5.1 tomorrow, we need to look into why automatic publishing failed.

rafaelramalho19:
I've high-level tested everything and seemed okay with the update, but I might have missed something due to my lack of knowledge of the app 😅

Also updated some dependencies with patches since we're already updating one.

Closes #1500


package.json
```diff
@@ -80,12 +80,12 @@
     "electron-updater": "^4.3.1",
     "fix-path": "^3.0.0",
     "fs-extra": "^9.0.0",
+    "go-ipfs-dep": "^0.5.1",
```
lidel, line 83: remove ^ please, we don't want to ship versions of go-ipfs that we did not explicitly test :)

rafaelramalho19: Nice catch! I missed that 😅

lidel:
The CI error on macOS was due to unrelated issue, fixed in https://github.com/ipfs-shipyard/ipfs-desktop/pull/1503

PR #1556, opened 2020-06-22
fix: windows auto-update feature when selecting install for all users

### LINKED ISSUE #1514: Auto update not working on Windows when installed globally for all users
- OS: Windows 10
 - Version of IPFS Desktop [e.g. 0.11.2]

**Describe the bug**

image

**To Reproduce**
2. Launch the application

**Expected behavior**

rafaelramalho19:
Fixes #1514

### DIFF
--- a/src/auto-updater/index.js
+++ b/src/auto-updater/index.js
@@ -1,4 +1,4 @@
-const { shell } = require('electron')
+const { app, shell } = require('electron')
 const { autoUpdater } = require('electron-updater')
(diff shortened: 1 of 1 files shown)


rafaelramalho19:
> don't we also need to add code that sets `_installOnQuit` ? (set it to `true` in `update-downloaded` etc)?

Yes! Sorry about that, was juggling between my macbook & desktop and ended up screwing the one I wanted to push 😅

PR #1619, opened 2020-08-27
Replace v0.11.4 with v0.12.2 as the latest release

aeruower:
The title says it all.


aeruower:
I just noticed this line:
> ⚠ Please note that this version is not stable yet and might change. Also, Linux support is still experimental and it might not work on every desktop environment. Please [file an issue](https://github.com/ipfs-shipyard/ipfs-desktop/issues/new) if you find a bug.

If it applies to v0.12.2, shouldn't you mark it as pre-release than release (as shown [here](https://docs.github.com/en/github/administering-a-repository/about-releases))?

jessicaschilling:
@aeruower Thanks so much for catching *and* updating this. You're also correct that the "not stable" language is out of date; I've removed this in https://github.com/ipfs-shipyard/ipfs-desktop/pull/1619/commits/acc3872cf80cb82585d255047cc9fcdc1eaa23b2.

Issue #1631, opened 2020-09-08
Typo in IPFS config property Discovery.MDNS.Enabled


olizilla:
`Enabled` should start with a capital `E` in `Discovery.MDNS.Enabled`, as per https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#discoverymdnsenabled

Currently ipfs-desktop sets the property name to be `enabled`
https://github.com/ipfs-shipyard/ipfs-desktop/blob/c387c4f4b2acd2cdc616cc326a9bbfb092d1c3da/src/daemon/config.js#L55

as discovered by @daviddias 

![image(3)](https://user-images.githubusercontent.com/58871/92505457-053a9780-f1fc-11ea-952a-5499c878b4be.png)

jessicaschilling:
@rafaelramalho19 and/or @lidel, please have a look if we want to prioritize this above the weekly triage. Thanks!

rafaelramalho19:
Thanks for opening the issue @olizilla, will address this now

PR #1632, opened 2020-09-08
fix: uppercase Discovery.MDNS.enabled in default config

rafaelramalho19:
Fixes https://github.com/ipfs-shipyard/ipfs-desktop/issues/1631

### DIFF
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -52,7 +52,7 @@ function applyDefaults (ipfsd) {
 
   config.Discovery = config.Discovery || {}
   config.Discovery.MDNS = config.Discovery.MDNS || {}
-  config.Discovery.MDNS.enabled = true
+  config.Discovery.MDNS.Enabled = true
 
   writeConfigFile(ipfsd, config)
 }
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -57,6 +57,48 @@ function applyDefaults (ipfsd) {
   writeConfigFile(ipfsd, config)
 }
 
+// Apply one-time updates to the config of IPFS node.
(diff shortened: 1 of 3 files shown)


olizilla:
@lidel do you think we should have ipfs desktop check for the incorrect key in the config, and remove it where it finds it?

lidel:
Good question. Right now we only control the defaults applied on clean install.

There is a need to have some generic way of migrating go-ipfs configs of _existing_ users for things like `ConnMgr` limits, so to make this and future changes easier, I've added `migrateConfig` function () which checks revision and applies one-time migration, if needed. 

@rafaelramalho19 @olizilla mind reviewing my changes?

lidel (approved):
LGTM, but because I've added `migrateConfig` would appreciate second set of eyes before merge

src/daemon/config.js
```diff
@@ -52,11 +52,53 @@ function applyDefaults (ipfsd) {
...
+    config = readConfigFile(ipfsd)
+  } catch (err) {
+    // This is a best effort check, dont blow up here, that should happen else where.
+    logger.error(`[daemon] migrateConfig: error reading config file: ${err.message || err}`)
+    return
+  }
+
+  // Cleanup https://github.com/ipfs-shipyard/ipfs-desktop/issues/1631
+  if (config.Discovery && config.Discovery.MDNS && config.Discovery.MDNS.enabled) {
+    config.Discovery.MDNS.Enabled = config.Discovery.MDNS.Enabled || true
```
hacdias, line 83: Isn't this `true` everytime?

lidel: no, user could remove it or explicitly set it to false to disable noise in LAN

lidel:
Merging to unblock #1638

PR #1653, opened 2020-09-20
fix: package.json, package-lock.json & .snyk to reduce vulnerabilities

Kirobi92:
The following vulnerabilities are fixed with an upgrade:
- https://snyk.io/vuln/SNYK-JS-I18NEXT-575536
- https://snyk.io/vuln/SNYK-JS-I18NEXT-585930

The following vulnerabilities are fixed with a Snyk patch:
- https://snyk.io/vuln/SNYK-JS-LODASH-567746

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,10 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.19.0
+ignore: {}
+# patches apply the minimum changes required to fix a vulnerability
+patch:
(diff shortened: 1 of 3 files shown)


lidel:
Thank you for submitting the PR, but I'm not able to merge it, because it introduces vendor-specific patching via a thirdparty tool.

In IPFS repos we usually bump dependencies before each release, and have github dependency check enabled, so adding Snyk would only increase noise and bring the opposite effect.

Issue #1661, opened 2020-10-02
Smoke-testing IPFS Desktop v0.13.0 


lidel:
> We now have new a brand CI/CD pipeline responsible for building IPFS Desktop (#1657, #1659).
> Before we ship v0.13, let's have more eyes on platform-specific packages.

## Help us by doing a quick test of IPFS Desktop v0.13.0 

If you have spare time between now and Monday (~5min):

1. Install IPFS Desktop using package listed below
2. Go to _Files_ and _Import_ a folder with some cat pictures
3. You should be able to see successfully imported files in your _Files_ screen
4. Comment on this issue with package name and if it went smoothly or  if you had any issues.

## RC2 build @ f2f34bfde7e46998ff915de5630da2914e7225e2

https://bafybeiamxdw3qlecv2mryeojfn46kv32t3nt2a6zajm6wotjadznrtvutm.ipfs.dweb.link/ipfs-desktop-0.13.0-rc2

## RC1 build @ 2cfd2f4f7a417fa371e543416a79e5a40971d137

<details>
  <summary>Click to expand</summary>

### macOS

- [x] [IPFS-Desktop-0.13.0.dmg](https://bafybeianbwepwyb5jglfbtxy74spitvhku5vjgwclemb4rolasw256hfda.ipfs.dweb.link/ipfs-desktop-0.13.0-rc1/IPFS-Desktop-0.13.0.dmg) @jessicaschilling  @olizilla 

### Windows 

- [x] [IPFS-Desktop-Setup-0.13.0.exe](https://bafybeianbwepwyb5jglfbtxy74spitvhku5vjgwclemb4rolasw256hfda.ipfs.dweb.link/ipfs-desktop-0.13.0-rc1/IPFS-Desktop-Setup-0.13.0.exe) @johnnymatthews @lidel 

### Linux

If you happen to run Linux, make sure to include the name of distro and Desktop Environment you are running.

- [x] [ipfs-desktop-0.13.0-linux-x64.tar.xz](https://bafybeianbwepwyb5jglfbtxy74spitvhku5vjgwclemb4rolasw256hfda.ipfs.dweb.link/ipfs-desktop-0.13.0-rc1/ipfs-desktop-0.13.0-linux-x64.tar.xz) (@lidel: unpacked standalone version runs as expected)
[...]

PR #1666, opened 2020-10-06
chore: webui v2.11.3

lidel:
https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.11.3

This aims to unblock #1661 by including fix from  https://github.com/ipfs-shipyard/ipfs-webui/pull/1660


lidel:
Smoke test on linux looks fine, merging to create RC2

Issue #1677, opened 2020-10-07
Auto-generated report: errors occurred upon startup


NeoGe-IPFSMain:
Please describe what you were doing when this error happened.

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.13.0
- **Electron Version**: 9.3.2
- **Chrome Version**: 83.0.4103.122

**Error**

```
TypeError: shell.openItem is not a function
    at click (C:\Users\Neo\AppData\Local\Programs\IPFS Desktop\resources\app.asar\src\tray.js:141:32)
    at MenuItem.click (electron/js2c/browser_init.js:65:1624)
    at Function.executeCommand (electron/js2c/browser_init.js:73:1102)
```

jessicaschilling:
Thanks, @NeoGe-IPFSMain. Are you able to successfully run Desktop despite this error?

NeoGe-IPFSMain:
Hi, @jessicaschilling. No, I am not able to run it with this error.

jessicaschilling:
@NeoGe-IPFSMain Thanks for the context -- please accept apology for having to run an earlier version in the meantime. @lidel and @rafaelramalho19 -- can you please have a look? Related to https://github.com/ipfs-shipyard/ipfs-desktop/issues/1570 perhaps?

lidel:
@NeoGe-IPFSMain which menu item are you clicking before this error occurs?

NeoGe-IPFSMain:
@jessicaschilling No worries! 

@lidel This is happening without any interactions. I double-clicked the IPFS Desktop shortcut on my desktop, but it refused to open. Right-click the icon on my taskbar, it shows "IPFS has Errored". 

It's been a while since the last time I used IPFS Desktop. I tried the restart and reinstall (not clean), this error remains.

jessicaschilling:
@NeoGe-IPFSMain We're actively triaging this right now - are you OK in the meantime using the CLI, or the CLI as daemon plus webui.ipfs.io?

NeoGe-IPFSMain:
@jessicaschilling CLI works fine. IPFS Desktop would be up running without an issue if I manually use any shell to run the IPFS daemon. But once I close the shell, IPFS Desktop will stop running and reporting the error.

jessicaschilling:
@NeoGe-IPFSMain Thanks for the details - just wanted to make sure you weren't completely blocked while we worked on a fix.

[...]

Issue #1676, opened 2020-10-07
Advanced → "Open [..] directory" cause fatal crash


jessicaschilling:
I'm repeatedly killing Desktop when trying to select the following options under `Advanced`:
- `Open Logs Directory`
- `Open Repository Directory`
- `Open Configuration File`

**Specifications**

- **OS**: darwin
- **IPFS Desktop Version**: 0.13.0
- **Electron Version**: 9.3.2
- **Chrome Version**: 83.0.4103.122

**Error**

```
TypeError: shell.openItem is not a function
    at click (/Applications/IPFS Desktop.app/Contents/Resources/app.asar/src/tray.js:141:32)
    at MenuItem.click (electron/js2c/browser_init.js:65:1624)
    at Function.executeCommand (electron/js2c/browser_init.js:73:1102)
```

lidel:
I can confirm the same error on Linux.


**Specifications**

- **OS**: linux
- **IPFS Desktop Version**: 0.13.0
- **Electron Version**: 9.3.2
- **Chrome Version**: 83.0.4103.122

**Error**

```
TypeError: shell.openItem is not a function
    at click (/home/lidel/tmp/test-13/ipfs-desktop-0.13.0-rc3/ipfs-desktop-0.13.0-linux-x64/resources/app.asar/src/tray.js:141:32)
    at MenuItem.click (electron/js2c/browser_init.js:65:1624)
    at Function.executeCommand (electron/js2c/browser_init.js:73:1102)
```

lidel:
Potential duplicate: https://github.com/ipfs-shipyard/ipfs-desktop/issues/1677

PR #1679, opened 2020-10-07
fix: windows auto-update

### LINKED ISSUE #1570: Auto-update broken on Windows
This issue is directly related to issue #1514 .  Not sure why this was closed before verifying that it worked, because, when I initiated the auto update to ipfs-desktop 12.0, the exact same issue occurred once again. Program was uninstalled, not re-installed, and the installation files were located in the %USERPROFILE%\Appdata\local\ipfs-desktop-updater. I did notice that there were 2 executable files in the directory though (I don't recall if there were last time): one named "updater.exe" and one that was the actual install file for the program (IPFS-Desktop-Setup-0.12.0.exe). Interestingly, while both were the same size, (83,195 kb), I noticed that the hash values were different, so they are not the same file? I don't know. What I do know is that the process is still broken and the potential cause of the issue which I mentioned in the 2nd of the 2 potential causes I listed remains unchanged (update-info.json "isAdminRightsRequired" value is still euqal to false, and as I stated the first time, you need admin permissions to install an application to the program files directory, which is where ipfs desktop is installed when you install it for all users). Maybe I am an outlier in this mindset, but I am of the belief that issues should be confirmed resolved before closing them.Auto update not working on Windows when installed globally for all users

 - OS: Windows 10
 - Version of IPFS Desktop [e.g. 0.12.0]

**Describe the bug**

image

**To Reproduce**
2. Launch the application

**Expected behavior**

lidel:
This PR aims to close #1570
Context: https://github.com/ipfs-shipyard/ipfs-desktop/issues/1570#issuecomment-705081315

## WIP

- [x] revert #1556 to reduce noise
- [x] figure out local setup for testing autoupdate on windows
  - We won't know for sure, unless we  test in production, but we don't need to use this app
  - I created [sandbox electron app](https://github.com/lidel/electron-updater-altschmerz), which uses electron-build config similar to one in ipfs-desktop and basically the same CI setup
  - Next step is to copy autoupdate code from ipfs-desktop to that app, and see how it behaves on Windows.
- [x] see if the  latest `electron-builder` fixed the original problem upstream
  - no, still broken https://github.com/ipfs-shipyard/ipfs-desktop/pull/1679#issuecomment-705569228
- [x] if still broken: fix windows autoupdate 
  - fix: https://github.com/ipfs-shipyard/ipfs-desktop/pull/1679#issuecomment-705630973
- [x] confirm it works when installed for single user
- [x] confirm it works when installed for every user

cc @jessicaschilling @rafaelramalho19

### DIFF
--- a/assets/build/nsis.nsh
+++ b/assets/build/nsis.nsh
@@ -38,3 +38,14 @@ ManifestDPIAware true
   DeleteRegKey SHELL_CONTEXT "Software\Classes\ipfs"
   DeleteRegKey SHELL_CONTEXT "Software\Classes\dweb"
 !macroend
+
+!macro customInit
+    # https://github.com/ipfs-shipyard/ipfs-desktop/pull/1679#issuecomment-705630973
+    ${if} $installMode == "all"
+        ${IfNot} ${UAC_IsAdmin}
+            ShowWindow $HWNDPARENT ${SW_HIDE}
+            !insertmacro UAC_RunElevated
+            Quit
+        ${endif}
+    ${endif}
+!macroend
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -43,6 +43,7 @@ nsis:
   oneClick: false
   warningsAsErrors: false
   perMachine: false
+  allowElevation: true
   allowToChangeInstallationDirectory: true
 
 linux:
--- a/src/auto-updater/index.js
+++ b/src/auto-updater/index.js
@@ -1,4 +1,4 @@
-const { app, shell } = require('electron')
+const { shell } = require('electron')
 const { autoUpdater } = require('electron-updater')
 const i18n = require('i18next')
 const logger = require('../common/logger')
@@ -7,26 +7,10 @@ const { showDialog } = require('../dialogs')
 const quitAndInstall = require('./quit-and-install')
 
 let feedback = false
-let installOnQuit = false
 
 function setup (ctx) {
   autoUpdater.autoDownload = false
-  autoUpdater.autoInstallOnAppQuit = false
-
-  /**
-   * this replaces the autoInstallOnAppQuit feature of autoUpdater, which causes the app
-   * to uninstall itself if it is installed for all users on a windows system.
-   *
-   * More info: https://github.com/ipfs-shipyard/ipfs-desktop/issues/1514
-   * Should be removed once https://github.com/electron-userland/electron-builder/issues/4815 is resolved.
-   */
-  app.once('before-quit', ev => {
-    if (installOnQuit) {
-      ev.preventDefault()
-      installOnQuit = false
-      autoUpdater.quitAndInstall(false, false)
-    }
-  })
+  autoUpdater.autoInstallOnAppQuit = true
 
   autoUpdater.on('error', err => {
     logger.error(`[updater] ${err.toString()}`)
@@ -97,8 +81,6 @@ function setup (ctx) {
 
   autoUpdater.on('update-downloaded', ({ version }) => {
     logger.info('[updater] update downloaded')
-
-    installOnQuit = true
 
     const doIt = () => {
       setImmediate(() => {


jessicaschilling:
@lidel - is this related to https://github.com/ipfs-shipyard/ipfs-desktop/issues/1677?

lidel:
@jessicaschilling unclear. that error could be a side effect of autoupdate issue, or a duplicate of #1676 – asked for more information.

lidel:
Closing the shop for today, but quick update:

I looked into setting up local update server, but it effectively becomes _Rube Goldberg machine_ and I don't believe it is a reliable way of testing and debugging this. To be sure we test the right thing, we literally need to test _in production_.

For that reason I created a [test app](https://github.com/lidel/electron-updater-altschmerz) with CI and electron config similar to Desktop. 

I'll copy relevant autoupdate code from Desktop tomorrow, so it is identical, and then we have setup to run some end-to-end tests.

lidel:
I tested the simplified autoupdate provided by `electron-updater` on windows and it works fine when app is installed for current user, but fails when "all users" option is selected during install:

> Capture-user-picker


This means original bug is still present in `electron-updater`. 
I'll now check if a viable fix exists, if not, we have to either disable "all users" option or detect it and disable automatic install, asking user for manual update instead"

lidel:
Disclaimer: I am not a Windows person, so this may be obvious stuff, but took me a while to understand all the moving pieces. Having the [altschmerz test app](https://github.com/lidel/electron-updater-altschmerz) helped a lot.

I believe the crux of the problem was that we use custom NSIS settings, namely:

```yaml
nsis:
  oneClick: false
  warningsAsErrors: false
  perMachine: false
  allowToChangeInstallationDirectory: true
```

By default we install per user (`perMachine: false`), but we also give user an option to change that via the screen from my previous comment (`oneClick: false`).

We don't want to require admin all the time, because users should be able to install IPFS on machines where they don't have Admin (schools, libraries). That is why `isAdminRightsRequired` is `false` in the update manifest (as described in https://github.com/ipfs-shipyard/ipfs-desktop/issues/1570#issue-652502677). 

`electron-builder` has a bunch of default macros for NSIS, and some higher level flags that control some Windows-specific behaviours, for example, [`electron-builder/**/nsis/NsisTarget.ts#L438-L444`](https://github.com/electron-userland/electron-builder/blob/v22.8.1/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L438-L444) sets:

```js
    if (options.perMachine === true) {
      defines.INSTALL_MODE_PER_ALL_USERS = null
    }

    if (!oneClick || options.perMachine === true) {
      defines.INSTALL_MODE_PER_ALL_USERS_REQUIRED = null
    }
```

`INSTALL_MODE_PER_ALL_USERS` and `INSTALL_MODE_PER_ALL_USERS_REQUIRED` are dynamic flags that control various things, one of them is deciding if installer will ask windows for elevated privileges before doing its thing.

In our case, `INSTALL_MODE_PER_ALL_USERS` is ON and `INSTALL_MODE_PER_ALL_USERS_REQUIRED` is OFF, and `UAC_RunElevated` is only used in `_REQUIRED` scenario – see [`electron-builder/**/nsis/multiUser.nsh#L52-L62`](https://github.com/electron-userland/electron-builder/blob/b8d6905582b28a9b36f11308f9a72ebc32dc2c86/packages/app-builder-lib/templates/nsis/multiUser.nsh#L52-L62)


**TLDR** the fix here is to 
- (a) fix upstream 
- (b) extend `assets/build/nsis.nsh` with `customInit` script that requests `UAC_RunElevated` when `${if} $installMode == "all"` (independent of `INSTALL_MODE_PER_ALL_USERS_REQUIRED`)

I'll prepare (b) first, as (a) will take long time before the fix lands upstream, and I still don't trust there won't be any regressions, as we had so many in the past, unfortunately.

lidel:
Implemented the fix in  0bf120d – ok to merge and ship patch release?

rafaelramalho19:
Tested and looks good, but we can only really find out on the next version... so 🚢 it !

rafaelramalho19 (approved):

witcher112:
Hi guys!

I encountered this PR while resolving some of the UAC issues with electron-builder auto updater.

Just wanted to let you know that there might be an alternative solution - modifying the access rights after the installation, to grant full access permissions to the non-admin users (to be honest not my idea, I just observed what Steam does) so **no administrative rights are required for the updater**.

Here's the snippet for custom NSIS include:
```
!macro customInstall
  AccessControl::GrantOnFile "$INSTDIR" "(S-1-5-32-545)" "FullAccess"
  pop $0
  AccessControl::GrantOnFile "$INSTDIR" "(S-1-5-18)" "FullAccess"
  pop $0
!macroend
```

_`S-1-5-32-545` is a SID for Users group_
_`S-1-5-18` is a SID for SYSTEM group_

Note that you need to link [AccessControl](https://nsis.sourceforge.io/AccessControl_plug-in) plugin to NSIS either through `!addplugindir` or by putting it into `build/x86-unicode` ([source](https://www.electron.build/configuration/nsis#custom-nsis-script)).

Cheers and good luck with your project!

lidel:
@witcher112 thank you, useful to know there is an alternative approach in case current one stops working.

Maybe I misunderstood the meaning of `GrantOnFile` (or just lacking Windows background), but.. isn't granting write access to non-admin users effectively a decrease in security? 

IIUC one user could overwrite binary that other users run.

PR #1682, opened 2020-10-08
fix(windows): autoInstallOnAppQuit

### LINKED ISSUE #1570: Auto-update broken on Windows

### LINKED ISSUE #1630: IPFS Desktop stuck at "IPFS is Starting" on Windows 10
Please describe what you were doing when this error happened.

Just trying to install and use it.  Stuck on "IPFS is Starting".  Tried rebooting.  Tried reinstalling.  Tried updating.  When I tried to update, it crashed with the following:

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.12.2
- **Electron Version**: 8.2.5
- **Chrome Version**: 80.0.3987.165

**Error**

```
TypeError: ctx.checkForUpdates is not a function
    at click (C:\Program Files\IPFS Desktop\resources\app.asar\src\tray.js:198:30)
    at MenuItem.click (electron/js2c/browser_init.js:1501:9)
    at Function.executeCommand (electron/js2c/browser_init.js:1762:13)
```

lidel:
https://github.com/lidel/electron-updater-altschmerz/issues/1 demonstrated that `autoInstallOnAppQuit` works fine on Windows, but we had a bunch of code responsible for doing it manually, probably due to macOS not being supported natively by `autoInstallOnAppQuit`.

Docs for `electron-updater` 4.3.5 state:
>  Screenshot_2020-10-09 Auto Update - electron-builder

Due to this, the fix from https://github.com/ipfs-shipyard/ipfs-desktop/pull/1679 was not applied correctly: macOS-specific handling executed first and broke updates on Windows.

This PR ensures that macOS keeps using the old upgrade paths, but Windows and AppImage leverage upstream code for applying upgrades.

Closes #1570 and most likely closes #1630  too.


lidel:
Sadly this last mile is specific to how our codebase and won't fix old versions.

To properly test this we need to tag v0.13.1 and v0.13.2 and do quick smoke test by installing v0.13.1 and testing if it upgrades fine to v0.13.2. 

Autoupdate on Windows does not require signing, so I'll try to do the above test in a fork first.

lidel:
Ok, tested this on Windows with my fork and both per user and per all users upgraded correctly on Windows.
There should be no change to behaviour on macOS.

src/daemon/index.js
```diff
@@ -109,12 +109,8 @@ module.exports = async function (ctx) {
 
   ipcMain.on('ipfsConfigChanged', restartIpfs)
 
-  app.on('before-quit', async e => {
-    if (ipfsd) {
-      e.preventDefault()
-      await stopIpfs()
-      app.quit()
```
lidel, line 116: not sure what was the purpose of this, but it seemed to interfere with `autoInstallOnAppQuit
` on Windows

src/auto-updater/index.js
```diff
@@ -4,13 +4,17 @@ const i18n = require('i18next')
 const logger = require('../common/logger')
 const { notify } = require('../common/notify')
 const { showDialog } = require('../dialogs')
-const quitAndInstall = require('./quit-and-install')
+const macQuitAndInstall = require('./macos-quit-and-install')
+const macOS = process.platform === 'darwin'
```
hacdias, line 8: Just a nitpick: we already have this constants defined on [`../common/consts`](https://github.com/ipfs-shipyard/ipfs-desktop/blob/master/src/common/consts.js).

hacdias (approved):
LGTM!

PR #1689, opened 2020-10-09
feat: screenshot improvements

lidel:
This ensures the CID is properly passed to copied link, and that filename is preserved when a single screen setup is used (produces single screenshot). Switched to reliable gateway, as share.ipfs.io is not maintained.

We also avoid low level pinning  (MFS is enough) and ensure the produced CID is the same as on CLI (avoiding problems described in https://github.com/ipfs-shipyard/ipfs-webui/issues/676).

### DIFF
--- a/src/take-screenshot.js
+++ b/src/take-screenshot.js
@@ -19,9 +19,11 @@ async function makeScreenshotDir (ipfs) {
   }
 }
 
-async function onSucess (ipfs, launchWebUI, path, img) {
-  const stats = await ipfs.files.stat(path)
-  const url = `https://share.ipfs.io/#/${stats.hash}`
+async function onSuccess (ipfs, launchWebUI, path, img) {
+  // preserve filename if single file is shared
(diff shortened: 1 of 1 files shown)


jessicaschilling (approved):
Confirmed this works for me on Mac, including generating two separate files (Screen 1.png and Screen 2.png) for my multi-screen setup.

One nit: the directory names aren't super user-friendly, probably out of scope for this issue but just wanted to note it.
image

lidel:
@jessicaschilling good catch. In  IPFS Companion we use `/ipfs-companion-imports/%Y-%M-%D_%h%m%s/` and numbers are padded with zeroes. Seconds are kinda needed if someone takes multiple screenshots one after another, just for sorting purposes. 

I've ported what we do in Companion as a starting point, let me know if anything should be tweaked:

> 

- month was wrong (had to +1, because JS lol)
- added padding and removed spaces for better lexicographical sorting and better-looking share URL

jessicaschilling:
@lidel Thank you! This sounds good.

PR #1690, opened 2020-10-09
fix: use correct config path during development

lidel:
This PR makes instance started via `npm start` read `package.json` and that makes Electron to use config from `~/.config/IPFS Desktop` instead of `~/.config/Electron`.

Goal here is to have the same behavior in dev as  in packaged production version, reducing the surface for human error (eg. I was removing `~/.config/IPFS Desktop` between tests while `~/.config/Electron` remained with broken config)

@hacdias @olizilla cc fyi (I remember this  being an issue in past, this seems to be the fix)

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,8 +173,6 @@ The configuration file and logs are located on:
 
 For quick access to this folders, just right-click on your tray icon and then 'Logs Directory' or 'Configuration File', depending on what you want.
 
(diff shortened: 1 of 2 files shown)


jessicaschilling (approved):

hacdias (approved):
If it fixes it, I'm all for it!

rafaelramalho19:
I think you repeated yourself with "~/.config/IPFS Desktop instead of ~/.config/IPFS Desktop", did you mean:
~/.config/IPFS Desktop instead of ~/.config/Electron ?

lidel:
@rafaelramalho19 correct – i fixed PR description, thanks!

rafaelramalho19 (approved):

PR #1691, opened 2020-10-09
fix: OS integrations for quick file import

### LINKED ISSUE #1684: "Add to IPFS" OS-level integrations are broken
https://github.com/ipfs-shipyard/ipfs-desktop#easy-add-to-ipfs states:

> #### Easy add to IPFS
> 
> You can easily add files and folders to IPFS:
> 
> - On Windows, you can right click on files to add them to IPFS through IPFS Desktop.
> - On macOS, you can drag and drop them to the tray icon.

Unfortunately, it seems that functionality is broken in v0.13.0:

```
error: [add to ipfs] started Error: TypeError: ipfsd.api.add(...) is not a function or its return value is not async iterable
at module.exports (C:\Program Files\IPFS Desktop\resources\app.asar\src\add-to-ipfs.js:103:14)
```

Most likely got broken due to JS AP change (async iterator + introduction of `addAll`)

@rafaelramalho19 as you have access to both macOS and Windows, mind taking a look?

lidel:
This PR refactors code to fix bugs caused by [js api changes](https://blog.ipfs.io/2020-02-01-async-await-refactor/).

The sneaky one: the way we pass template changed in a way that failed silently:
https://github.com/ipfs/js-ipfs/blob/8cb8c73037e44894d756b70f344b3282463206f9/packages/interface-ipfs-core/src/object/new.js#L33-L40
so instead of unixfs, we were patching untyped object, which gateway was unable to render.

@rafaelramalho19 
I've pushed 1 line fix to your branch and created this PR to include our fixes.

### How to test

Same code is used for different variant on each OS, so we need to test both.

#### macOS

- [x] drag&drop multiple files or a file and a directory to desktop's icon

#### Windows

- [x] right click on a file and select "Add to IPFS" from context menu

Closes #1684

### DIFF
--- a/src/add-to-ipfs.js
+++ b/src/add-to-ipfs.js
@@ -32,7 +32,7 @@ async function makeShareableObject (ipfs, results) {
     return results[0]
   }
 
-  let baseCID = await ipfs.object.new('unixfs-dir')
+  let baseCID = await ipfs.object.new({ template: 'unixfs-dir' })
 
   for (const { cid, path, size } of results) {
     baseCID = (await ipfs.object.patch.addLink(baseCID, {
--- a/src/add-to-ipfs.js
+++ b/src/add-to-ipfs.js
@@ -87,11 +87,7 @@ module.exports = async function ({ getIpfsd, launchWebUI }, files) {
 
   await Promise.all(files.map(async file => {
     try {
-      let result = null
-      for await (const res of ipfsd.api.add(globSource(file, { recursive: true }))) {
-        result = res
-      }
-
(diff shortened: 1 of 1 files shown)


jessicaschilling (approved):
Successfully dropped onto the menubar icon to import:
- One file
- Multiple files
- A file plus a directory

rafaelramalho19:
Looks great! Thanks for picking up from my first bug fix 😄 

🚢  itttttttt

jessicaschilling (approved):
Confirmed import worked and clipboard link displayed successfully for:

- One file
- Two files
- A file plus a directory

Issue #1683, opened 2020-10-08
electron-updater errors on Linux


NukeManDan:
- OS: Linux
 - Version of IPFS Desktop v0.13.0

**Describe the bug**
I get a non-fatal error when trying to update as described below. 

Ported from [this web-ui issue](https://github.com/ipfs-shipyard/ipfs-webui/issues/1665)

**To Reproduce**
1. right click tray icon 
2. About...
3. Check for updates....
4. Error (see attached files)

**Expected behavior**
Updating the GUI and CLI :pray:

**Additional context**
Also can confirm #1676 - I can't open logs dir from the tray icon. Crashes the app.

<details>
  <summary>terminal output on startup, try update via stem tray icon</summary>

```
2020-10-08T23:27:41.997Z info: [meta] logs can be found on /home/dan/.config/IPFS Desktop
Error: ENOTDIR: not a directory, mkdir '/opt/IPFS Desktop/resources/app.asar/node_modules/countly-sdk-nodejs/data'
    at Object.mkdirSync (fs.js:823:3)
    at Object.Countly.init (/opt/IPFS Desktop/resources/app.asar/node_modules/countly-sdk-nodejs/lib/countly.js:157:24)
    at module.exports (/opt/IPFS Desktop/resources/app.asar/src/analytics.js:6:11)
    at run (/opt/IPFS Desktop/resources/app.asar/src/index.js:65:11)
Error: ENOENT, node_modules/countly-sdk-nodejs/data/__cly_id.json not found in /opt/IPFS Desktop/resources/app.asar
    at createError (electron/js2c/asar.js:111:17)
    at Object.func [as openSync] (electron/js2c/asar.js:146:27)
    at Object.writeFileSync (fs.js:1265:35)
    at storeSet (/opt/IPFS Desktop/resources/app.asar/node_modules/countly-sdk-nodejs/lib/countly.js:1337:16)
    at Object.Countly.init (/opt/IPFS Desktop/resources/app.asar/node_modules/countly-sdk-nodejs/lib/countly.js:171:21)
    at module.exports (/opt/IPFS Desktop/resources/app.asar/src/analytics.js:6:11)
    at run (/opt/IPFS Desktop/resources/app.asar/src/index.js:65:11)
2020-10-08T23:27:42.800Z info: [web ui] window ready
2020-10-08T23:27:42.801Z info: [web ui] navigate to /
2020-10-08T23:27:42.855Z info: [tray] starting
[...]

Issue #1274, opened 2019-12-03
Cryptic error and unreliable daemon on linux after update to 0.10.2


NerdyKyogre:
As of the latest version on my ubuntu 18.04.3 machine, starting IPFS Desktop at login results in a gray icon with the status "IPFS has errored." The daemon has failed to start, which may be related. As soon as I manually run ipfs daemon from a terminal, IPFS desktop seems to work happily. However, the gray icon and error message persist, and restarting IPFS Desktop has no effect. I should mention that I am running the deepin desktop environment and have installed IPFS Desktop as a snap.

bertrandfalguiere:
Idem for Ubuntu 19.10, installed from sources.

hacdias:
This is really weird because we did not update/change any dependencies or code related to the daemon logic. Looking at the changelog, I see I made a change I shouldn't have [https://github.com/ipfs-shipyard/ipfs-desktop/compare/v0.10.1...v0.10.2#diff-bc78997f8a1c5c7e9e703382e4341c96R150](done): where is `on` it should be `off`. However, none of those changes should make the daemon fail to start.

Can I ask you to check the logs, please? I'm pretty sure that if there was an error, it must've been logged. Please check the last section of the README to know where to find the logs on each OS! :)

bertrandfalguiere:
This issue might be older than that, unfortunately. I remember having a similar problem months ago, but that was before I understood IPFS well enough. I haven't used it in the interim, so I wouldn't know how it was since. 
The issue of @NerdyKyogre reminded me of it. 

I can check on older versions, but I don't know how to install versions from sources.

I'll check the logs in the next(ish) days.

NerdyKyogre:
Sorry for late reply, spam filter decided this was spam.

In the logs, the error that seems to be causing this issue is the following:
2020-01-06T16:38:15.284Z error: [ipfsd] start daemon Error: Error: mkdir /home/colef/snap/ipfs-desktop/5/.ipfs: no such file or directory
Note that ~/snap/ipfs-desktop/5 does not exist on my system at all. This looks like some weirdness with how snaps work that I don't fully understand.

This is spammed every time IPFS desktop tried to start up and the daemon failed. I'm currently getting around this by starting the daemon and desktop combined with XDG autostart, which functions, but still shows "IPFS has errored."

There's also this from the updater, but I don't think that's related:
[...]

Issue #928, opened 2019-05-14
Fresh install on Fedora 30 won't start


jkonecny12:
Fresh installation on Fedora 30 desktop. The application won't even start.

This error message is the only thing I'm able to find:

```
cat ~/.config/IPFS\ Desktop/error.log                                                                                                   
2019-05-14T12:51:25.841Z error: ENOENT: no such file or directory, open '/opt/IPFS Desktop/resources/app-update.yml'
```

If you need more info, you can ask.

BTW the file is really missing.

EDIT: I forgot to mention that I used rpm package for installation.

jkonecny12:
Adding another log:
```
cat ~/.config/IPFS\ Desktop/combined.log
2019-05-14T12:51:25.565Z info: [meta] logs can be found on /home/jkonecny/.config/IPFS Desktop
2019-05-14T12:51:25.841Z error: ENOENT: no such file or directory, open '/opt/IPFS Desktop/resources/app-update.yml'
2019-05-14T12:51:26.154Z info: [web ui] window ready
2019-05-14T12:51:26.154Z info: [tray] starting
2019-05-14T12:51:26.187Z info: [tray] started
2019-05-14T12:51:26.187Z info: [ipfsd] starting daemon
2019-05-14T12:51:28.193Z info: [ipfsd] daemon started
2019-05-14T12:51:28.210Z info: [hash download] shortcut disabled
2019-05-14T12:51:28.210Z info: [screenshot] shortcut disabled
2019-05-14T12:51:28.211Z info: [launch on startup] disabled
```

stiangrindvoll:
This is not unique to Fedora 30, also after installing the deb package on ubuntu 18.04

hacdias:
Close in favour of #1153.

PR #1698, opened 2020-10-13
fix: autoupdate only on supported platforms

### LINKED ISSUE #1671: Error: ENOENT: no such file or directory, open '/opt/IPFS Desktop/resources/app-update.yml
- OS: Ubuntu 20.04
 - Version of IPFS Desktop : v0.12.1

**Describe the bug**
When using a .appImage, the app is updated as expected.
If I use an other install method (.tar.xz, .deb, .snap), the app fails to update. I think the problem would be the same on Fedora for .rpm installs.

**To Reproduce**
1. Install IPFS Desktop v0.12.1
2. Click on About/Check for updates...
3. See error :

error.log shows: 
`2020-10-07T14:54:12.831Z error: [updater] Error: ENOENT: no such file or directory, open '/opt/IPFS Desktop/resources/app-update.yml'`

Error appears right after start-up, or when manually checking for updates.

**Expected behavior**
IPFS Desktop should upgrade to v0.12.2 or newer, like when launched via a .appImage.

**Additional context**
I'm guessing that the `app-update.yml` mentioned in the error is generated by `electron-builder` only for AppImage by default when publishing (see [here](https://www.electron.build/auto-update.html)).

**Proposed solutions**
If possible, electron-builder should enable the auto-update of a .deb or .snap file. I'm not sure that this is possible as it is the package managers' job.
If not possible, .deb and .snap (and .rpm) should be added to their native package manager's repositories: APT and Snapstore (and RPM). Might need some work, though. Maybe the recent move to GitHub Action can help? (https://github.com/ipfs-shipyard/ipfs-desktop/pull/1657) 
If not possible, PL could host releases, and IPFS Desklearntop could be fetched via IPFS. :) IPFS Desktop would check for updates via a ping to ipfs.io, or by subscribing to a pubsub channel where PL announce new releases. Would be awesome, but probably a lot of code to write and maintain. This was already partially proposed by @lidel here: https://github.com/ipfs-shipyard/ipfs-desktop/issues/1263#issuecomment-558416578 (all but the pubsub channel)
If not possible, https://github.com/ipfs-shipyard/ipfs-desktop/issues/1488 should add some UI candies on update failure and warn that non-AppImage users need to monitor Github to check when a release is available.

**Logs**
error.log:
`2020-10-07T15:52:48.610Z error: [updater] Error: ENOENT: no such file or directory, open '/opt/IPFS Desktop/resources/app-update.yml'`

lidel:
This PR:

- disables autoupdate on platforms other than macOS, Windows and AppImage, solving problem described in #1671
- on platforms that do not have built-in autoupdate we assume a third-party package manager is used
  - The `About → Check for Update..` menu item is still present, but instead of built-in check it opens https://github.com/ipfs-shipyard/ipfs-desktop/releases/latest

Closes #1671 and also historical  #1683 / #1274 / #928

### DIFF
--- a/src/auto-updater/index.js
+++ b/src/auto-updater/index.js
@@ -5,7 +5,13 @@ const logger = require('../common/logger')
 const { notify } = require('../common/notify')
 const { showDialog } = require('../dialogs')
 const macQuitAndInstall = require('./macos-quit-and-install')
-const { IS_MAC } = require('../common/consts')
+const { IS_MAC, IS_WIN, IS_APPIMAGE } = require('../common/consts')
+
+function isAutoUpdateSupported () {
+  // atm only macOS, windows and AppImage builds support autoupdate mechanism,
+  // everything else needs to be updated manually or via a third-party package manager
+  return IS_MAC || IS_WIN || IS_APPIMAGE
+}
 
 let feedback = false
 
@@ -137,6 +143,12 @@ module.exports = async function (ctx) {
     }
     return
   }
+  if (!isAutoUpdateSupported()) {
+    ctx.manualCheckForUpdates = () => {
+      shell.openExternal('https://github.com/ipfs-shipyard/ipfs-desktop/releases/latest')
+    }
+    return
+  }
 
   setup(ctx)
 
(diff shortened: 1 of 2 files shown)


jessicaschilling (approved):
Nice call on adding the releases page as a trapdoor.

DONALDSULLENDER (approved):
love it

rafaelramalho19 (approved):

rafaelramalho19 (approved):

jessicaschilling:
Is this ready to merge?

PR #1708, opened 2020-11-06
chore: switch actions to v2

lidel:
@olizilla any concerns regarding switching to v2?


olizilla:
None! i've found v2 setup-node to be much faster.

drumenbassik (approved):

duc811997:
[github-recovery-codes.txt](https://github.com/ipfs-shipyard/ipfs-desktop/files/5863409/github-recovery-codes.txt)

PR #1727, opened 2021-01-05
Fix Homebrew instructions

rmorey:
the syntax `brew cask install <pkg>` is now disabled. as per: https://github.com/Homebrew/discussions/discussions/340#discussioncomment-232364


jessicaschilling (approved):
Thanks for keeping this up to date!

Issue #1728, opened 2021-01-08
Text coding error in Chinese


aihy:
Text coding error in Chinese

1. Click on Download
![image](https://user-images.githubusercontent.com/16832365/103969249-7ab17700-51a0-11eb-87e9-718363789d3b.png)
2. See two buttons text error
![image](https://user-images.githubusercontent.com/16832365/103969287-8d2bb080-51a0-11eb-93c7-d55382b5bbd8.png)

at macOS Big Sur 11.1
ipfs-desktop 0.13.2

jessicaschilling:
https://github.com/ipfs-shipyard/ipfs-desktop/pull/1729 should fix this.

PR #1729, opened 2021-01-08
Explicitly specify content-type in prompt template

jessicaschilling:
Closes https://github.com/ipfs-shipyard/ipfs-desktop/issues/1728.

This PR explicitly specifies `utf-8` content-type in the template for the "Download..." dialog so non-roman languages don't blow out.

Tested in Mac Big Sur **only**; would appreciate a look in Linux/Windows.

### Before
image

### After
image

### DIFF
--- a/src/dialogs/prompt/template.js
+++ b/src/dialogs/prompt/template.js
@@ -1,7 +1,7 @@
 module.exports = ({ pallette, message, defaultValue, buttons, id }) => (`<!DOCTYPE html>
 <html>
   <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+    <meta charset="utf-8" /> 
   </head>
   <body>
     <p>${message}</p>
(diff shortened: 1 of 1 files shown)


aihy:
Thank you for fixing it!

src/dialogs/prompt/template.js
```diff
@@ -1,5 +1,8 @@
 module.exports = ({ pallette, message, defaultValue, buttons, id }) => (`<!DOCTYPE html>
 <html>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
```
lidel, line 4: There is a shorter notation in HTML5: 
```suggestion
    <meta charset="utf-8" /> 
```

lidel (approved):
Tested on Linux, works as expected :+1: 
Feel free to merge with or without comment below :-)

src/dialogs/prompt/template.js
jessicaschilling: Clearly I'm old 😊  Thanks for pointing that out!

PR #1735, opened 2021-01-19
feat: enable pubsub via settings menu

andrew:
Part of https://github.com/ipfs-shipyard/ipfs-desktop/issues/1647

Screenshot 2021-01-19 at 13 48 38

Adds a menu preference option to enable pubsub which adds a flag to the ipfsd-ctl options for startup.

Question: should this prompt a restart, given that the setting won't take effect until ipfs has been been restarted?

Bug: I'm getting the enabled/disabled message logged twice when clicking on the menu item, not quite sure why :
```
$ npm start

> ipfs-desktop@0.13.2 start
> cross-env NODE_ENV=development electron .

2021-01-19T16:19:06.988Z info: [meta] logs can be found on /Users/andrewnesbitt/Library/Application Support/IPFS Desktop
2021-01-19T16:19:10.560Z info: [web ui] window ready
2021-01-19T16:19:10.560Z info: [tray] starting
2021-01-19T16:19:10.607Z info: [tray] started
2021-01-19T16:19:10.608Z info: [ipfsd] start daemon STARTED
2021-01-19T16:19:10.937Z info: [daemon] removing api file
2021-01-19T16:19:11.954Z info: [ipfsd] start daemon FINISHED 1.3460042808055879s
2021-01-19T16:19:11.957Z info: [launch on startup] unavailable during development
2021-01-19T16:19:11.972Z info: [pubsub] disabled
2021-01-19T16:19:11.975Z info: [ipfs on path] no action taken
2021-01-19T16:19:17.770Z info: [pubsub] enabled
2021-01-19T16:19:17.791Z info: [pubsub] enabled
2021-01-19T16:22:36.969Z info: [pubsub] disabled
2021-01-19T16:22:37.000Z info: [pubsub] disabled
```

Issue #1647, opened 2020-09-16
Menu items for enabling pubsub and automatic GC


lidel:
I was looking into starting go-ipfs with additional flags: [`--enable-pubsub-experiment`](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#ipfs-pubsub) (pubsub) and  [`--enable-namesys-pubsub`](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#ipns-pubsub) (IPNS over pubsub) and the main concern I've found was use of resources:

> In order to enable pubsub by default in go-ipfs, we need to find some way for pubsub to not take up a bunch of resources when not in-use.
> – https://github.com/libp2p/go-libp2p-pubsub/issues/332

So if we want pubsub and fast IPNS enabled, we need to pay the cost of higher resource utilization.
We could measure it and decide if its worth the added value, or prioritize a fix.

**Updated:** or at the very least, we could give user UI for enabling that with one click

**Updated again:** See mockup for placement of menu items for pubsub and automatic GC (note this also includes not-yet-implemented "Ask when opening" menu item, likely to land first in https://github.com/ipfs-shipyard/ipfs-desktop/issues/1646)
![image](https://user-images.githubusercontent.com/1507828/93796230-c5aa7d00-fbf7-11ea-98ac-de36c3018e4b.png)

@aschmahmann are there any other concerns that I don't see?

@autonome FYSA this also impacts go-ipfs in Brave, especially the speed of `ipns://`

shaikh-shahid:
I think if we have a UI based switch to turn on/off the flags such as Pubsub and GC etc then that would be better for end users as well. I am currently working on a framework that does require IPFS in the backend and needs the PUBSUB to be enabled. Since the framework also consists of demos for general public as well, it's difficult for users to find the configuration and update the flags.

lidel:
@shaikh-shahid that is very useful feedback, thank you.

I think we could avoid enabling pubsub by default by adding  on/off checkbox settings for pubsub (and automatic GC) to the Desktop app menu. Toggling would automatically restart the deamon with the new setting.

jessicaschilling:
Note: Added mockup of menu item placement in [original comment](https://github.com/ipfs-shipyard/ipfs-desktop/issues/1647#issue-702650415). 

Unless @aschmahmann or @autonome have objections, I believe this is ready to work.

andrew:
[...]

PR #1739, opened 2021-01-20
feat: enable ipns over pubsub via settings menu

andrew:
Part of #1647

Very similar approach to #1735

Screenshot 2021-01-20 at 16 08 21

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -224,6 +224,7 @@
     "launchOnStartup": "Launch at Login",
     "openWebUIAtLaunch": "Open Web UI at Launch",
     "pubsub": "Enable PubSub",
+    "namesysPubsub": "Enable IPNS over PubSub",
     "ipfsCommandLineTools": "Command Line Tools",
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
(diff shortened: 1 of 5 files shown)


jessicaschilling:
Hey @andrew, sorry to swoop in on your PR, but as part of this work do you mind changing the menu item `npm on IPFS` to `Use npm on IPFS`? That way everything in the Experiments menu has a verb (and avoids the awkward lowercase of "npm").

Apologies that I didn't catch this under #1735 as it would have been more germane there 😟

andrew:
@jessicaschilling would "Enable npm on IPFS" be more inline with the other experiment options?

jessicaschilling:
@andrew Was leaning toward "Use" for the sake of avoiding visual repetition and because it _might_ be a slightly better word in this case -- i.e. you're not really enabling npm, you're just using it via IPFS delivery. But that's definitely not gospel.

andrew:
Alternatively, you are installing npm on ipfs, which gets used outside of ipfs desktop `Install npm on IPFS`?

jessicaschilling:
@andrew Good point. Yes please 😊

andrew:
@jessicaschilling broke that change out into a separate PR: https://github.com/ipfs-shipyard/ipfs-desktop/pull/1742

lidel (approved):
Thanks! 

(Moved experiments to pre-existing `experiments.` section of the config)

PR #1740, opened 2021-01-20
feat: disable/enable gc via settings menu

### LINKED ISSUE #1647: Menu items for enabling pubsub and automatic GC

andrew:
Closes #1647 (now that #1735 has been merged)

Very similar approach to #1735

Screenshot 2021-01-20 at 16 21 01

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -224,7 +224,7 @@
     "launchOnStartup": "Launch at Login",
     "openWebUIAtLaunch": "Open Web UI at Launch",
     "pubsub": "Enable PubSub",
-    "automaticGc": "Automatic Garbage Collection",
+    "automaticGC": "Automatic Garbage Collection",
     "ipfsCommandLineTools": "Command Line Tools",
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -224,6 +224,7 @@
     "launchOnStartup": "Launch at Login",
     "openWebUIAtLaunch": "Open Web UI at Launch",
     "pubsub": "Enable PubSub",
+    "automaticGc": "Automatic Garbage Collection",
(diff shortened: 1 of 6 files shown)


jessicaschilling:
Hey @andrew, sorry to swoop in on your PR, but as part of this work do you mind changing the menu item `npm on IPFS` to `Use npm on IPFS`? That way everything in the Experiments menu has a verb (and avoids the awkward lowercase of "npm").

Apologies that I didn't catch this under #1735 as it would have been more germane there 😟

jessicaschilling:
@lidel, for your review 😊

lidel (approved):
Thanks @andrew! 


Made small tweaks:
- added migration setting default state to active for new and pre-existing config
- removed keysize (not needed since go-ipfs 0.7.0 uses ED25519 keys by default)

PR #1742, opened 2021-01-22
Tweak title of npm on ipfs preference

andrew:
As requested by @jessicaschilling in https://github.com/ipfs-shipyard/ipfs-desktop/pull/1739#issuecomment-763893741

Screenshot 2021-01-22 at 13 50 36

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -228,7 +228,7 @@
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
     "experiments": "Experiments",
-    "npmOnIpfs": "npm on IPFS"
+    "npmOnIpfs": "Install npm on IPFS"
   },
   "setCustomIpfsBinaryConfirmation": {
     "title": "Custom IPFS binary",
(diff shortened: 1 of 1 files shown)


jessicaschilling (approved):
LGTM but I'll defer to @lidel for OK/merging 😊

assets/locales/en.json
```diff
@@ -228,7 +228,7 @@
     "takeScreenshotShortcut": "Global Screenshot Shortcut",
     "downloadHashShortcut": "Global Download Shortcut",
     "experiments": "Experiments",
-    "npmOnIpfs": "npm on IPFS"
+    "npmOnIpfs": "Install npm on IPFS"
```
lidel, line 231:
```suggestion
    "npmOnIpfs": "Enable npm on IPFS"
```

lidel (changes requested):
Given this is a toggle checkbox which installs/uninstalls, and other Experiments start with _Enable_, can we do the same? :thinking:

assets/locales/en.json
jessicaschilling: I'd specifically asked to use "Enable" since it's a little more accurate, and avoids visual repetition. If we' use "Enable" for each one, we might as well remove the word "Enable" altogether.

lidel: Let's go with Enable here, and after I merge all other PRs we can clean it up later.

PR #1757, opened 2021-02-10
fix: default Open at Launch to true

### LINKED ISSUE #1741: Default "Open Web UI at Launch" to on for all OSes
Suggesting that we change the default setting of `Preferences`->`Open Web UI at Launch' to `true` across all OSes. Right now we make this so for Linux, since some desktop envs don't have a tray for our "cube menu" to live in. However, this doesn't account for users on other OSes who may have extensions for decluttering system menus/trays, and thus have no initial "way in" to Desktop upon install. Ref https://github.com/ipfs-shipyard/ipfs-desktop/issues/1733#issuecomment-763559838.

Acknowledging that this means that users who have both `Launch at Login` and `Open Web UI at Launch` selected will see Web UI every time they start their machine. However, this is consistent with how many, many other apps approach this scenario.

sutartmelson:
Fixes #1741 by defaulting `openWebUIAtLaunch` to `true`.

This change is particularly useful for users who have extensions that declutter their system/tray which may initially hinder access to the Desktop app.

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -7,7 +7,6 @@ const os = require('os')
 const openExternal = require('./open-external')
 const logger = require('../common/logger')
 const store = require('../common/store')
-const { IS_MAC, IS_WIN } = require('../common/consts')
 const dock = require('../utils/dock')
 const { VERSION, ELECTRON_VERSION } = require('../common/consts')
 const createToggler = require('../utils/create-toggler')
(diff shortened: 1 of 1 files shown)


jessicaschilling:
Thanks for contributing this, @sutartmelson!

@lidel - do you mind a code review? Thank you!

lidel (approved):
Thank you @sutartmelson, lgtm.
Merging. We will ship this when go-ipfs 0.8.0 lands.

Issue #1488, opened 2020-05-04
Improve UX when updating


lidel:
Right now user can manually force update via `About` → `Check for Updates`
but they don't get any feedback until its done or fails. Downloading updates may take time on slower connection, producing experience where nothing happens for a long long time.

We could:
- at least do what we do for GC, and set status in menu to `Running (Fetching Update)` (https://github.com/ipfs-shipyard/ipfs-desktop/pull/1420#issuecomment-615884957).
- (optional) additionally, we could pass progress to webui when opened (inspiration: https://github.com/unipept/unipept-desktop/pull/35#issue-384231781)

jessicaschilling:
Both these options are clear and co sister to with how we are already doing things. I’d suggest both.

jessicaschilling:
Menu text changes:
- Gray out "Check for Updates..." and replace text with "Checking for Updates"
- Replace menu status with yellow dot (same as GC) and "Running (Checking for Updates)"

Snackbar feedback text:
- If failed, failure-specific message (what types of failure other than "could not download" do we have?) with link to troubleshoot: at minimum, dist page
- Is there a need for an additional "taking a long time" message or does this just turn into failure on timeout?

jessicaschilling:
Is this something worth addressing in our current batch of small Desktop changes? Not sure if either @lidel or @rafaelramalho19 would have time for it?

lidel:
Closed by https://github.com/ipfs/ipfs-desktop/pull/1758

PR #1758, opened 2021-02-12
Improve updating ux

sutartmelson:
This change provides simple enhancements to UX when updating the client, addressing feedback brought up in #1488.

Normal state:
CheckForUpdates

While checking for or downloading update:
CheckingForUpdates

This does not address everything mentioned in #1488, specifically snackbar notifications. I couldn't find an obvious way to do that without changing ipfs-webui, but i'm also very new to these projects so please advise if there's something I could have done.


jessicaschilling:
Thanks, @sutartmelson! Nice little fix.
@lidel - can you please review when you get the chance? 🙏

lidel (approved):
Thank you @sutartmelson  :heart: :raised_hands: 

The code inside `updateMenu` gets crowded, but still acceptable, we can refactor it in separate PR when something else is added.

We will include this in Desktop release with go-ipfs 0.8.0

Issue #1312, opened 2020-01-03
Show number of discovered nodes in the taskbar menu under activity indicator


rslnx:
That is important to see not only that IPFS is running but if it is online i.e. connected to any other nodes. Would be nice to have the number of nodes connected in the menu under the "running" indicator.

jessicaschilling:
Thanks, @shoce, and apologies for the delayed response. Might this be something you'd be willing to PR?

jessicaschilling:
@hacdias Wanted to get your opinion on this if you had a minute ...
- Seems reasonable that we could include this either in the "IPFS is Running" indicator in the menu, or maybe in the "Nodes" menu item that brings up Web UI's Nodes screen (though the former probably makes more sense) ... 
   - But how often would we want to update this (assuming we get this number from Web UI)?
   - Are there performance implications?
- Also, this approach would be visually different than Companion, which indicates the number of nodes in the cube icon itself. The little bit of digging I did (and admittedly regarding MacOS only) indicated that trying a similar approach of having the menubar icon change dynamically to reflect the number of nodes was impractical or maybe even impossible. Is that true, in your opinion?

hacdias:
Hey!

> But how often would we want to update this (assuming we get this number from Web UI)?

That only depends on how often you need the peers count to update. Maybe we could use the same value as IPFS Companion (which by just looking at it seems to be 1 second).

> Are there performance implications?

Great question. It would need us to rebuild the entire menu every time. Since it is not visible most of the time, I'm not sure if there are any practical implications.

> Is that true, in your opinion?

What I remember reading the last time I searched about it is that it is hard, really really hard. So I'd not go for it.

lidel:
Could this be in the tooltip that is displayed when user hovers over the icon (without clicking on it)?
I think that would be more useful than adding it to menu which requires a click.

jessicaschilling:
What about doing both? Just from generalized user testing, someone's tendency to actually use hover hints is a pretty personal trait.

jessicaschilling:
https://github.com/ipfs-shipyard/ipfs-desktop/pull/1599 stubs out how this would appear in both locations:

![image](https://user-images.githubusercontent.com/1507828/89353637-09f7b300-d674-11ea-9c90-54920221e87c.png)

![image](https://user-images.githubusercontent.com/1507828/89353655-1714a200-d674-11ea-9b23-c9e5bbd9737f.png)

[...]

PR #1759, opened 2021-02-14
feat: Show discovered peer count

sutartmelson:
Addressing feature request from #1312. The peer count is retrieved by polling the ipsf daemon every 60 seconds. Upon peer count retrieval, the menu is rebuilt and the tray tooltip is updated.

Updating more frequently is almost certainly overkill as this isn't meant to give detailed information about your connectivity, so up to date accuracy isn't necessary. Polling for the peer count in response to opening the menu or hovering over the tray (thereby only polling when explicitly needed) is too slow and the menu can't be rebuild in place once its opened (AFAIK).

If electron updates its API to allow dynamically editing menu labels, this feature could be improved.

In tray:
PeerCountInTray

In menu:
peerCountInMenu

Tray and menu when IPFS isn't running:
NoPeersIPFSStopped


lidel (changes requested):
I really want to avoid constant polling, even if it is once a minute.

Would it be possible to change this PR to remove polling?

I was looking at tray APIs while working on #1762  and perhaps we could refresh cached peer count via this event:
- https://www.electronjs.org/docs/api/tray#event-mouse-move-macos-windows 

The above API is not available on Linux, but it should not be an issue in practice:
The value is not real-time anyway, so we could simply display the last cached value + refresh cache asynchronously every time  the menu is rendered (in addition to  `mouse-move` event).

sutartmelson:
@lidel, Great idea, if polling is to be avoided (which is understandable), displaying the last cached value and fetching a new one at menu open is a good compromise. The displayed peer count isn't perfectly up to date, but I don't think it needs to be. I also updated the PR description accordingly. Hopefully this last change is a step in the right direction. Thanks!

lidel (changes requested):
I rebased this PR and fixed pluralization of `peerCount` label in the menu.

Hoped to include it in the release but was unable to make it work on Linux at all: turns out when "app indicator" tray backend is used on Linux, the `click` events are ignored, so we have no way to trigger peerCount refresh before menu is displayed. 

## TODO


- [ ] figure out what to do on Linux (find a way to trigger this, disable, or poll)
- [ ] we seem to have two "Peers" items now – one on the top, and then "Peers" link for opening Peers screen in webui. I think we want to deduplicate this (eg. remove "Peers" and make one with count clickable and open Peers screen win webui)

AlexxNica:
@lidel, how is this showing on Linux atm? If it's already disabled I think we can just leave it as is for Linux until we have a fix so we're not holding this issue just for that.

For the second to-do item, I'm with you that only one "Peers" should remain.

PR #1767, opened 2021-02-23
fix(mac): force-refresh file list after import

### LINKED ISSUE #1763: (Mac) Force Files screen refresh when clicking on drag/drop success notification
_Note: Only tested on Mac, so not sure about impact on other OSes._

To reproduce problem:
- Drag/drop multiple items onto the MacOS menubar icon
- Wait until the "click here to view your files" notification pops up
- Click it and get sent to the Files screen, which doesn't yet reflect the existence of the new files even though they're there. Refreshing or clicking away/back from Files screen shows the files

To fix:
- Force page reload of Files screen upon clicking notification

Note:
- Not sure if there's any consideration that needs to be made for single-file drag/drop -- these currently work fine, because clicking the notification takes you directly to the file, not the whole Files screen.

lidel:
Closes #1763

### TODO

- [x] get someone with macOS to confirm this fixes the thing (after CI builds the DMG)

### DIFF
--- a/src/add-to-ipfs.js
+++ b/src/add-to-ipfs.js
@@ -45,7 +45,7 @@ async function makeShareableObject (ipfs, results) {
   return { cid: baseCID, path: '' }
 }
 
-function sendNotification (failures, successes, launch, path) {
+function sendNotification (failures, successes, launchWebUI, path) {
   let link, title, body, fn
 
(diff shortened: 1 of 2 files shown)


jessicaschilling (approved):
Confirmed it works as intended on Big Sur 11.2.1.

Issue #1774, opened 2021-02-25
ipfs not found after snap updated to latest


apunkt:
- OS:  Linux Ubuntu 20.04
 - Version of IPFS Desktop 0.14.0

No connection to IPFS-API

**To Reproduce**
1. Start IPFS-Desktop

**Expected behavior**
It connects successfully to the IPFS daemin

**Additional context**
Automatic update failed and leads to permission error and missing binaries for ipfs daemong

`2021-01-25T08:28:02.610Z error: [updater] Error: ENOENT: no such file or directory, open '/snap/ipfs-desktop/23/resources/app-update.yml'
2021-01-25T08:41:00.092Z error: [ipfsd] start daemon SyntaxError: /home/andreas/snap/ipfs-desktop/23/.ipfs/config: Unexpected token ] in JSON at position 455
    at JSON.parse (<anonymous>)
    at Object.readFileSync (/snap/ipfs-desktop/23/resources/app.asar/node_modules/fs-extra/node_modules/jsonfile/index.js:52:17)
    at readConfigFile (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/config.js:29:13)
    at checkPorts (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/config.js:217:18)
    at module.exports (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/daemon.js:74:24)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
    at async startIpfs (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/index.js:54:15)
    at async restartIpfs (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/index.js:102:5)
    at async Object.restartIpfs (/snap/ipfs-desktop/23/resources/app.asar/src/daemon/index.js:33:5)
2021-01-25T10:54:30.741Z error: [updater] Error: ENOENT: no such file or directory, open '/snap/ipfs-desktop/23/resources/app-update.yml'
2021-01-26T07:59:27.512Z error: [updater] Error: ENOENT: no such file or directory, open '/snap/ipfs-desktop/23/resources/app-update.yml'
2021-01-27T08:05:30.749Z error: [updater] Error: ENOENT: no such file or directory, open '/snap/ipfs-desktop/23/resources/app-update.yml'
[...]

PR #1776, opened 2021-03-02
docs: promote AppImage for Linux

lidel:
AppImage is the only one that does autoupdates on Linux without a third-party package manager.
Pointing less sophisticated users at it may be beneficial, 
as they avoid issues with things like Snap which does not look reliable these days (https://github.com/ipfs-shipyard/ipfs-desktop/issues/1774#issuecomment-787210407)

cc @johnnymatthews  @olizilla @jessicaschilling

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,12 @@ Use IPFS Desktop to get acquainted with IPFS without needing to touch the termin
 
 ### Quick-install shortcuts
 
-- **Mac:** [IPFS-Desktop-0.14.0.dmg](https://github.com/ipfs-shipyard/ipfs-desktop/releases/download/v0.14.0/IPFS-Desktop-0.14.0.dmg) or `brew install --cask ipfs`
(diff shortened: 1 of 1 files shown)


jessicaschilling (approved):
Works for me!

johnnymatthews:
Sounds good. Just to check, I can only see one `AppImage` file in the [releases](https://github.com/ipfs-shipyard/ipfs-desktop/releases): `ipfs-desktop-0.14.0-linux-x86_64.AppImage`. Are we planning to create an AMD64 specific `AppImage`, or are we just gonna stick with the one `x86_64` image?

Sorry, that was a lot of CPU architectures 🤦

lidel:
@johnnymatthews in the context of packages `x86_64` is the same as `AMD64` (some distributions like Gentoo use "amd64" because AMD was first to introduce 64bit x86, but its copatible with Intel)

PR #1808, opened 2021-04-25
docs: fixed scoop install

dzmitry-lahoda:


lidel (approved):
LGTM (extras is backed by [lukesampson/scoop-extras](https://github.com/lukesampson/scoop-extras/blob/06d725ddc09aedd47fdac93800ac5fc9047a8454/bucket/ipfs-desktop.json))

PR #1843, opened 2021-06-16
fix: print useful error on write fail

lidel:
This displays error details when write to a destination directory fails, and ensures stored files are contained within destination dir.
Should improve UX in situations where user picked read-only directory, run out of disk space etc. 

cc @schomatis

### DIFF
--- a/src/download-cid.js
+++ b/src/download-cid.js
@@ -18,8 +18,11 @@ const SHORTCUT = IS_MAC
   : 'CommandOrControl+Alt+D'
 
 async function saveFile (dir, file) {
-  const location = join(dir, file.path)
-  await fs.outputFile(location, file.content)
+  const destination = join(dir, file.path)
+  if (!destination.startsWith(dir)) {
+    throw new Error(`unable to create '${file.path}' outside of '${dir}'`)
+  }
+  await fs.outputFile(destination, file.content)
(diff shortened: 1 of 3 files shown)


src/download-cid.js
```diff
@@ -18,8 +18,11 @@ const SHORTCUT = IS_MAC
   : 'CommandOrControl+Alt+D'
 
 async function saveFile (dir, file) {
-  const location = join(dir, file.path)
-  await fs.outputFile(location, file.content)
+  const destination = join(dir, file.path)
+  if (!destination.startsWith(dir)) {
+    throw new Error(`unable to create '${file.path}' outside of '${dir}'`)
+  }
```
schomatis, line 24: Without knowing anything about anything in JS land: are UnixFS symlinks also under a similar check?

lidel: Right now, symlinks are ignored by the download function in ipfs-desktop, only files and directories are written to disk.

But even if re-created symlinks, same destination check would be applied (symlink could point at external path, but its own location would be limited to destination dir).

lidel: Just to be sure this won't become an issue in the future, added c4cfa325dec7b7e6862c5c1258f01e600e81457b

PR #1894, opened 2021-08-14
fix: use context bridge

hacdias:
Electron 12 introduced context isolation enabled by default. Thus, the `window` object that the preload script and the Web UI page have access to are different. Hence, Web UI cannot detect if it's running inside Desktop. ~Disabling Context Isolation fixes the issue.~ Using the new `contextBridge` function fixes the issue, as well as exposing the language update function through the global object.
	  
## How can you see it is not working properly in the current version?
	  
Open Settings, go to the Analytics part and check the events list. If you don't see `SCREENSHOT_TAKEN` (among others) is because of this.

License: MIT

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -29,7 +29,6 @@ const createWindow = () => {
     webPreferences: {
       preload: join(__dirname, 'preload.js'),
       webSecurity: false,
-      contextIsolation: false,
       allowRunningInsecureContent: false,
       enableRemoteModule: process.env.NODE_ENV === 'test', // https://github.com/electron-userland/spectron/pull/738#issuecomment-754810364
       nodeIntegration: process.env.NODE_ENV === 'test'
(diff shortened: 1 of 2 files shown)


lidel (changes requested):
I am afraid of exposing powerful electron apis to the browser process– better to avoid this if we can, especially if it only impacts desktop detection and metrics, and has no impact on actual functionality.

Would it be possible to keep `contextIsolation: false` and use `contextBridge.exposeInMainWorld` instead  ([docs](https://www.electronjs.org/docs/tutorial/context-isolation#context-isolation)) ?

hacdias:
@lidel okay, I will implement this, but I will also be required to make a few other changes on the Web UI to make sure all integrations work.

hacdias:
@lidel updated to use `contextBridge`. I also added a function `updateLanguage` that will be explicitly called by Web UI when updating the language (see https://github.com/ipfs/ipfs-webui/pull/1839). This way we do not need to wrap the `localStorage.setItem`.

lidel (approved):
Perfect!  LGTM.

Let's merge this and https://github.com/ipfs/ipfs-webui/pull/1839. I'll bump webui version in desktop before release, so everything keeps working as expected.

PR #1896, opened 2021-08-26
feat: go-ipfs 0.10.0

guseggert:


lidel (changes requested):
Looks good, waiting for final release before the merge.

lidel:
Note to self: 
- CI is flaky because  fetching webui from public gateway  fails  sometimes. 
- Retry tomorrow, if the issue is still present, add some fallback

lidel (approved):

PR #1899, opened 2021-09-02
chore: update ipfs deps

hacdias:
Depends on #1895.

License: MIT


hacdias:
@lidel after upgrading the IPFS-related dependencies, now I get a 403 error whenever we try to call `ipfs.id()` unless I add `*` to the `Access-Control-Allow-Origin` API's field in the daemon settings. This seems quite odd as we're not in a browser environment. Were there any security-related changes on the HTTP Client or ipfsd-ctl that may have caused this?

lidel (changes requested):
(I rebased this PR on top of `main` which already included all the other updates)

@hacdias no idea what changed – perhaps some `fetch` polyfill deep inside on dependencies? I've seen some dark arts happening around https://github.com/ipfs/js-ipfs-utils/pull/136 :ghost: 

Setting `Access-Control-Allow-Origin` is a security risk, we need to find a proper fix.

lidel:
@achingbrain does this look familiar? canything landed in js-ipfs-http-client (or its deps) that could change the way `fetch` works?

achingbrain:
@lidel nothing springs immediately to mind & we're running `ipfs-http-client` tests against `go-ipfs@0.9.1` in CI, same as the deps here.

ipfs/js-ipfs-utils#136 is just to stop using the undocumented `npm:some-module` dep version override style in favour of a tarball so codepen-style environments can use it.  We need node-fetch with https://github.com/node-fetch/node-fetch/pull/1172 applied otherwise stream errors that occur during pubsub subscriptions are swallowed silently 🙄.

`47.x.x` to `52.x.x` is quite a big jump though, could you step through each version to find out which one stopped working?

hacdias:
@achingbrain upgrading from 47.x to 48.x seems to create the issue. I wonder if it is related to https://github.com/ipfs/js-ipfs/pull/3275

hacdias:
@lidel seems that removing our old CORS logic for IPFS Desktop solves the issue successfully. Could you please take a look at this PR?

(PS: I could not run the tests locally - https://github.com/electron-userland/spectron/issues/929 - gotta work on #1937)

src/daemon/daemon.js
```diff
@@ -79,7 +79,7 @@ module.exports = async function (opts) {
     logger.info(`[daemon] PeerID is ${id}`)
     logger.info(`[daemon] Repo is at ${ipfsd.path}`)
   } catch (err) {
-    if (!err.message.includes('ECONNREFUSED')) {
+    if (!err.message.includes('ERR_CONNECTION_REFUSED')) {
```
lidel, line 82: Do you remember why this change? 
Did Chromium/Electron change the error code?

src/webui/index.js
```diff
@@ -69,16 +68,6 @@ const createWindow = () => {
...
-// Converts a Multiaddr to a valid value for Origin HTTP header
-const apiOrigin = (apiMultiaddr) => {
-  // Return opaque origin when there is no API yet
-  // https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque
-  if (!apiMultiaddr) return 'null'
-  // Return the Origin of HTTP API
-  const apiUri = toUri(apiMultiaddr, { assumeHttp: true })
-  return new URL(apiUri).origin
-}
-
```
lidel, line 81: :information_source: yep, this is no longer needed due to https://github.com/ipfs/go-ipfs-cmds/pull/201

src/webui/index.js
```diff
@@ -145,26 +134,12 @@ module.exports = async function (ctx) {
     })
   })
 
-  // Avoid setting CORS by acting like /webui loaded from API port
+  // Set user agent
   session.defaultSession.webRequest.onBeforeSendHeaders((details, callback) => {
-    details.requestHeaders.Origin = apiOrigin(apiAddress)
     details.requestHeaders['User-Agent'] = `ipfs-desktop/${VERSION} (Electron ${ELECTRON_VERSION})`
```
lidel, line 139: :information_source:  Presence of `Electron` here makes  https://github.com/ipfs/go-ipfs-cmds/pull/201/files#diff-9cd5b98b32603671d06faab6568b1be0d3f2790b8ee426270cb20cb91824997aR182 allow the requests.

lidel (approved):
LGTM, but will merge after I'm back in January :)

src/daemon/daemon.js
hacdias: Good question. I can't find anywhere that indicates that this was changed in either Node.js or Electron. However, this PR does not update Electron, only some IPFS dependencies. I remember updating Node.js a few weeks ago. Maybe that's it?

In addition, I notice that most google results for `ERR_CONNECTION_REFUSED` are quite recent, while for `ECONNREFUSED` are a few years old.

hacdias: I added the check for both instead.

hacdias:
macOS build failing with `Failed to upload app to Apple's notarization servers`.

PR #1901, opened 2021-09-04
feat: show devtools via DEBUG=ipfs-desktop in env

milahu:


lidel (approved):
Thanks! Looks useful.

FYI  devtools can be opened via keyboard shortcut (<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>i</kbd>) on-demand even without this :)

PR #1909, opened 2021-09-20
fix: detect unresponsiveness and restart window

### LINKED ISSUE #1865: Recover from unresponsive webui in webContents
`QmfFLsxJsviSncH5LGkwss53Vn2v6ebZYb3bJYtZ5P25HM`

this is the a sample file hash.

browse this file hash will freeze the window, and got high cpu usage.

and the log said: `error: [web ui] unresponsive: [object Object]`

hacdias:
Fixes #1865

@lidel Still working on this. I'm having issues making the window unresponsive. Could not reproduce #1865 and using `process.hang()` does not trigger a complete window freeze but only the page itself. That is not triggering the unresponsive event.

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -45,6 +45,27 @@ const createWindow = () => {
 
   window.webContents.on('unresponsive', event => {
     logger.error(`[web ui] unresponsive: ${event.toString()}`)
+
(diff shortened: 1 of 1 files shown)


hacdias:
Superseeded by #1952

PR #1914, opened 2021-10-01
chore: default --agent-version-suffix and --routing

lidel:
[go-ipfs 0.10.0 shipped](https://github.com/ipfs/go-ipfs/releases/tag/v0.10.0) with support for running  `ipfs daemon` with custom `--agent-version-suffix`  (https://github.com/ipfs/go-ipfs/pull/8419)

This PR adds `--agent-version-suffix=desktop` and cleans up the way we set `--routing`
IPFS Desktop nodes will now announce themselves as `go-ipfs/0.10.0/desktop`

### DIFF
--- a/src/common/store.js
+++ b/src/common/store.js
@@ -6,9 +6,10 @@ const defaults = {
     type: 'go',
     path: '',
     flags: [
+      '--agent-version-suffix=desktop',
       '--migrate',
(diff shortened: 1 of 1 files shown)


lidel:
Merging as this is not a controversial change + we can always back-away from this via autoupdate in future versions.

Issue #1893, opened 2021-08-12
Avoid running Renderer when WebUI is not shown


unbeatable-101:
**Is your feature request related to a problem? Please describe.**
Using electron takes up a lot of memory compared to other native programs.

**Describe the solution you'd like**
Write a native app that is only the menubar (e.g. Status, Files, and Peers would open your web browser).

**Describe alternatives you've considered**
Don't start IPFS Desktop Helper (Renderer) if `openWebUIAtLaunch` is set to `false`, rewrite the whole app to be native (Maybe use the bundled WKWebView?)

lidel:
@mvdan  created  a light rewrite without Electron at: https://github.com/ipfs-shipyard/go-ipfs-desktop
You can give it a spin and report issues/feature requests in that repo – it is unofficial (unsigned) for now, but we want to provide it as alternative to Electron at some point.

unbeatable-101:
> ps. created a light rewrite without Electron at: ipfs-shipyard/go-ipfs-desktop – give it a try and report issues/feature requests in that repo – it is unofficial (unsigned) for now, but we want to provide it as alternative to Electron at some point.

Oh nice, this is basically exactally what I wanted

cyberpunkrocker-zero:
+1 for Electron-less go-ipfs-desktop! I wanted to use ipfs-desktop on my ageing laptop, but it was using so much resources that it was impossible to keep it running it when the laptop was off AC. Go-ipfs-desktop is a perfect low-power option.

tinytb:
backlog grooming: we are considering and experimenting with alternatives to electron, e.g. [here is a basic attempt to use wails](https://github.com/SgtPooki/ipfs-desktop-wails) instead of electron. Closing this issue, as these efforts will be tracked elsewhere:
- https://github.com/ipfs/ipfs-desktop/issues/2184
- https://github.com/ipfs/ipfs-desktop/issues/867

Issue #1902, opened 2021-09-06
Unblock Electron Update


lidel:
We are 2 versions behind, waiting for https://github.com/electron-userland/spectron/releases to ship something compatible with the latest Electron.

hacdias:
I will analyse this.

lidel:
Even tho Electron 15 shipped, we can't update.
This is still blocked by the lack of [Spectron](https://www.electronjs.org/spectron) (tool we use for E2E tests) compatible with Electron 14.

lidel:
so.. spectron is dead: https://github.com/electron-userland/spectron/issues/1045 :skull: 

means we need to invest more work here, and rewrite e2e tests to use playwright:
- https://github.com/electron-userland/spectron/issues/896#issuecomment-860134329
- https://github.com/tanshuai/electron-playwright-e2e-test-quick-start

hacdias:
I will look into moving to playwright.

PR #1929, opened 2021-11-19
proof of concept: close Web UI window instead of hiding 

hacdias:
This is a proof of concept for #1893. In this PR, the Web UI window is only created on demand, i.e., when we launch some URL on the Web UI or if the user explicitly set `openWebUIAtLaunch` to `true`. When the window is closed, it is actually closed and not hidden.

Electron runs 4 processes: Electron, Electron Helper, Electron Helper (GPU) and Electron Helper (Renderer). Only the renderer process is killed by the window closing. The GPU process is still on.

Problems with this approach:
- OSes that do not have a menubar functionality cannot close IFPS Desktop.

@lidel what do you think?

### DIFF
--- a/src/second-instance.js
+++ b/src/second-instance.js
@@ -12,6 +12,6 @@ module.exports = async function (ctx) {
       return
     }
 
-    ctx.launchWebUI()
+    await ctx.launchWebUI()
   })
 }
--- a/src/take-screenshot.js
+++ b/src/take-screenshot.js
@@ -99,9 +99,11 @@ function handleScreenshot (ctx) {
   }
 }
 
-function takeScreenshot (ctx) {
-  const { webui } = ctx
+async function takeScreenshot (ctx) {
+  const { getWebUI } = ctx
   logger.info('[screenshot] taking screenshot')
+
+  const webui = await getWebUI()
   webui.webContents.send('screenshot')
 }
 
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -96,13 +96,6 @@ const createWindow = () => {
     store.set('window.height', dim[1])
   })
 
-  window.on('close', (event) => {
-    event.preventDefault()
-    window.hide()
-    dock.hide()
-    logger.info('[web ui] window hidden')
-  })
-
   app.on('before-quit', () => {
     // Makes sure the app quits even though we prevent
     // the closing of this window.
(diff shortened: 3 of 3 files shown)


lidel (changes requested):
My understanding is that nothing changes for OS that does not have a menubar functionality: closing the window keeps the daemon running in the background, just like it was before. 

I am supportive for this change, however there is a regression on the Status screen:

The bandwidth graph it no longer gets populated when the window does not exist, so the user does not have correct bandwidth utility usage (it gets "paused" until window is opened again).

To solve this,  ipfs-desktop would have to keep polling the API and storing stats, and when a window is created inject them into historical data ipfs-webui uses for plotting the graph.

Not sure how feasible  this is, thoughts?

hacdias:
> My understanding is that nothing changes for OS that does not have a menubar functionality: closing the window keeps the daemon running in the background, just like it was before.

You're right indeed.

> To solve this, ipfs-desktop would have to keep polling the API and storing stats, and when a window is created inject them into historical data ipfs-webui uses for plotting the graph.

I think it is possible to do. I will take a look later today!

hacdias:
@lidel I also want to add here that I don't think I can fix the E2E tests easily: Spectron assumes that no active windows means the app is closed and thus it cannot shut down the app. In addition, [Spectron is being deprecated](https://github.com/electron-userland/spectron/issues/1045). The most sensible alternative seems to be [Playwright](https://github.com/microsoft/playwright) however I wonder if it'd suffer from the same issue since Electron apps were not built to be... menu bar apps.

Do you think I should go ahead with collecting with metrics in the background? If we follow that route, we need to see what to do with the tests.

lidel:
I am afraid we don't have bandwidth to tackle those in parallel. 
We should solve spectron problem (tracked  in #1902) and when we have e2e test situation resolved, get back to this PR.

lidel:
This is no longer blocked – we fixed e2e tests in #1937.
@hacdias mind rebasing?

SgtPooki:
If the attempt for this PR is to speed up the application, we can probably disregard it. There are multiple changes in flight that would make this change unnecessary, or at least cumbersome to do in parallel with.

* #2114 (ctx is changing)
* #2125 - I've already got load time to an unobservable delay - load times to come with #2126
* #2126 - we need metrics showing the problem before we accept any PRs claiming to fix them.

hacdias:
@SgtPooki the main concern, mentioned in #1893, is actually the resources that the rendered process consumes in background. This PR would actually make the app slightly slower, as the Web UI Window would need to be completely rebuilt every time we open it. Right now, we just hide the window when the user closes, instead of actually closing it.

SgtPooki:
@hacdias I think after https://github.com/ipfs/ipfs-desktop/pull/2125, "rebuilding the webui" everytime becomes a non-issue. So I'm completely fine with this PR and can take this over in the future if it's not something you can finish up

SgtPooki:
closing this since there is a large rebase required and we don't have the bandwidth to focus on this right now

PR #1947, opened 2022-01-04
refactor: uninstall npm on ipfs

hacdias:
I took the liberty to take the deprecation of 'npm on ipfs' further by uninstalling it for users that still have it enabled. This feature was first deprecated in February 2021 (17578745addb6c665b0c3a3e50499d3967c8efb5).

If the user still has the option enabled, it is uninstalled ad they get a dialog like this:

image

Alternatively, we can also totally remove this feature instead. I would prefer this option as npm on IPFS is a package that can be easily uninstalled manually.

License: MIT


lidel (approved):
LGTM. In case someone still has this enabled, it will try to uninstall once, and if that fails for some reason, won't bother users.

PR #1948, opened 2022-01-04
refactor: remove deprecated "ipfs on PATH" feature

### LINKED ISSUE #1943: Windows: EPERM %userprofile%\.ipfs-desktop\IPFS_PATH
Please describe what you were doing when this error happened.

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.18.1
- **Electron Version**: 13.6.3
- **Chrome Version**: 91.0.4472.164

**Error**

```
Error: EPERM: operation not permitted, open 'C:\Users\tmpri\.ipfs-desktop\IPFS_PATH'
    at Object.openSync (fs.js:476:3)
    at Object.func [as openSync] (electron/js2c/asar_bundle.js:5:1846)
    at Object.writeFileSync (fs.js:1467:35)
    at Object.outputFileSync (C:\Program Files\IPFS Desktop\resources\app.asar\node_modules\fs-extra\lib\output\index.js:31:15)
    at writeIpfsPath (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\index.js:130:6)
    at startIpfs (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\index.js:50:7)
    at module.exports (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\index.js:116:9)
    at run (C:\Program Files\IPFS Desktop\resources\app.asar\src\index.js:74:11)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

### LINKED ISSUE #1969: [gui error report]
Please describe what you were doing when this error happened.

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.18.1
- **Electron Version**: 13.6.3
- **Chrome Version**: 91.0.4472.164

**Error**

```
Error: EPERM: operation not permitted, open 'C:\Users\benfl\.ipfs-desktop\IPFS_PATH'
    at Object.openSync (fs.js:476:3)
    at Object.func [as openSync] (electron/js2c/asar_bundle.js:5:1846)
    at Object.writeFileSync (fs.js:1467:35)
    at Object.outputFileSync (D:\Users\benfl\AppData\Local\Programs\IPFS Desktop\resources\app.asar\node_modules\fs-extra\lib\output\index.js:31:15)
    at writeIpfsPath (D:\Users\benfl\AppData\Local\Programs\IPFS Desktop\resources\app.asar\src\daemon\index.js:130:6)
    at startIpfs (D:\Users\benfl\AppData\Local\Programs\IPFS Desktop\resources\app.asar\src\daemon\index.js:50:7)
    at module.exports (D:\Users\benfl\AppData\Local\Programs\IPFS Desktop\resources\app.asar\src\daemon\index.js:116:9)
    at run (D:\Users\benfl\AppData\Local\Programs\IPFS Desktop\resources\app.asar\src\index.js:74:11)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

hacdias:
This PR (almost) completely removes the feature 'IPFS on PATH'. This feature was officially deprecated in #1768 and only stayed active for users that had it enabled.

Even while deprecated, this feature is creating problems (see #1943). This PR completely uninstalls IPFS on PATH for the users that still have it enabled and shows this little message:

image

Fixes #1943
Closes #1969

In a few versions, we should remove the uninstall code. I think it is wise to keep for now so that it reverts the changes that it made before.


src/ipfs-on-path/scripts/uninstall.js
```diff
@@ -1,5 +1,44 @@
...
+    fs.copySync(backup, dst)
+    console.log(`${backup} restored to ${dst}`)
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      console.log(`${backup} not found, skipping`)
+      return
+    }
+
+    checkErr(err)
+  }
```
lidel, line 35: This _feels_ like a footgun :see_no_evil: 
My thinking: it is possible that someone enabled IPFS on PATH long time ago, and the `ipfs.bak` binary is really, really  old. If that is the case for some of our users, then restoring this ancient binary would create a whole new class of bugs (incompatible repo and/or rpc api).

I think it is better to remove backup but don't touch `/usr/local/bin/ipfs` (if it is a symlink, let it be – lesser evil):

```suggestion
```

src/ipfs-on-path/index.js
```diff
@@ -1,123 +1,74 @@
...
 
-async function runWindows (script, { failSilently }) {
-  return new Promise(resolve => {
+async function uninstallWindows () {
+  return new Promise((resolve, reject) => {
     execFile('powershell.exe', [
       '-nop', '-exec', 'bypass',
       '-win', 'hidden', '-File',
-      join(__dirname, `scripts/${script}.ps1`).replace('app.asar', 'app.asar.unpacked')
+      join(__dirname, 'scripts/uninstall.ps1').replace('app.asar', 'app.asar.unpacked')
```
lidel, line 52: Similar note as one for `uninstall.js` – I am not sure if running the `uninstall.ps1` script is safe:

https://github.com/ipfs/ipfs-desktop/blob/b73c1cc4ee6bc8efa9afdce13fa11ee8bb038f0c/src/ipfs-on-path/scripts/uninstall.ps1#L1-L3

Where does the `bin-win` come from? Is it something created by electron-builder for Windows?

 If I understand correctly, it is impossible to distinguish between user who enabled in before February 2021, and users who manually added the binary to their PATH? 
This means we could break someone's setup.

Let's leave `PATH` as-is to minimize breakage of power users.

lidel (changes requested):
Thank you for taking time and cleaning this up @hacdias  :pray: 

I think we should minimize the amount of invasive operations
See inline comments for uninstall on windows and other platforms.

If we adjust, this should be safe to ship.

src/ipfs-on-path/index.js
hacdias: `bin-win` is given by _us_: https://github.com/ipfs/ipfs-desktop/tree/main/src/ipfs-on-path/scripts/bin-win

I don't think anyone would've added that to the `PATH` manually. At least, they shouldn't. This file is also removed on this PR. So... even if they have it in the `PATH`, it won't work.

This script was already here. They're not new scripts and they've worked until now. I was just planning on reusing them.

src/ipfs-on-path/index.js
```diff
@@ -1,123 +1,74 @@
...
-        ]
-      }) !== 0) {
-        // User canceled
-        return
-      }
-
-      return run('install')
+    try {
+      unlinkSync(join(app.getPath('home'), './.ipfs-desktop/IPFS_PATH').replace('app.asar', 'app.asar.unpacked'))
+      unlinkSync(join(app.getPath('home'), './.ipfs-desktop/IPFS_EXEC').replace('app.asar', 'app.asar.unpacked'))
```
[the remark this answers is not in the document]
hacdias: Honestly, I can't remember. I think that, for some reason `app.getPath` returns a path where it includes `app.asar` but when the app is installed, that compressed file gets unpacked. I don't remember the specifics about it, only that we needed this to make it work.

This code was already in our codebase, I just moved it to `index.js` to cleanup the "ipfs on PATH" related files.

See (from `main`):

https://github.com/ipfs/ipfs-desktop/blob/4b989b135901f585c752e93c0d706bd2df20e72b/src/ipfs-on-path/scripts/consts.js

https://github.com/ipfs/ipfs-desktop/blob/4b989b135901f585c752e93c0d706bd2df20e72b/src/daemon/index.js#L129-L135

https://github.com/ipfs/ipfs-desktop/blob/3b138478835117469a4dcd89fca59e3fb25e5e18/src/daemon/daemon.js#L30-L36

So we're just removing the files we don't need anymore.

src/ipfs-on-path/index.js
```diff
@@ -1,123 +1,74 @@
...
-      message: i18n.t('disableIpfsOnPath.message'),
+    logger.info('[ipfs on path] uninstalled')
+
+    showDialog({
+      title: 'Command Line Tools Uninstalled',
+      message: 'Command Line Tools via IPFS Desktop have been deprecated in February 2021. They have now been uninstalled. Please refer to https://docs.ipfs.io/install/command-line/ if you need to use ipfs from the command line.',
       buttons: [
-        i18n.t('disableIpfsOnPath.action'),
-        i18n.t('cancel')
+        i18n.t('close')
```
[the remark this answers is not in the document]
hacdias: Good idea. Done.

src/ipfs-on-path/scripts/uninstall.js
hacdias: Okay, I agree with you on this one. I'll merge this bit and so it'll only remove the backup and keep symlinked to IPFS Desktop's binary, which may cause the least damage.

Shouldn't we also remove the symlink too though? Because we say the feature was disabled, so it's a bit odd that we'd keep it.

src/ipfs-on-path/index.js
lidel: Thanks for explaining, sgtm!

lidel (approved):
Thanks! 
Merging, as I believe we should ship this.
It does its best to clean up any remaining cruft for legacy users, but does not impact new ones.

(I did not bother to add i18n, assuming CLI users don't mind one-time message in English, since our CLI is only in English anyway)

Issue #1951, opened 2022-01-09
IPFS just goes Blank screen when try to download...


ZeroCool22:
- OS: Windows 10 x64
 - Version of IPFS Desktop [0.18.1]

**Describe the bug**

https://streamable.com/9z175t

**To Reproduce**
1. Go to the file
2. Click on more
3. Scroll down to Download
4. Blank Screen after some seconds.

**Log files:**

[combined.log](https://github.com/ipfs/ipfs-desktop/files/7834484/combined.log)
[error.log](https://github.com/ipfs/ipfs-desktop/files/7834485/error.log)

hacdias:
Hi @ZeroCool22! Thanks for the bug report. I could not reproduce it, which makes me think it may be related to RAM memory. I see you have 32 GB of RAM and that it reaches ~60% usage before crashing. How large is the file?

ZeroCool22:
> Hi @ZeroCool22! Thanks for the bug report. I could not reproduce it, which makes me think it may be related to RAM memory. I see you have 32 GB of RAM and that it reaches ~60% usage before crashing. How large is the file? Could you also try checking the developers console when it happens (CTRL+SHIFT+I)?

Here you go:

https://streamable.com/u0kchy

ZeroCool22:
No help?

I really want to use IPFS... :(

lidel:
@ZeroCool22 thank you for recording the demo, it is very useful. 
We will fix this, but for now you need to use a workaround: avoid using the "Download" ("Descarga") action from menu, and click on "Local gateway" instead:

> ![2022-01-11_23-25](https://user-images.githubusercontent.com/157609/149031253-9a1f080f-a2ed-4335-8312-81e851d91713.png)

Let us know if this works with your 7GB file.

---

@hacdias my understanding of the problem is that  we need to change how "Download" context action of a single file is implemented on the Files screen. It reads correct URL from `doFilesDownloadLink` but  then downloads it using `downloadFile` function from [src/files/download-file.js](https://github.com/ipfs/ipfs-webui/blob/3910e86fd90e0a958338efe961219dfd2ed04375/src/files/download-file.js#L18), which buffers the entire thing in memory to construct a `blob:`.  

I believe creating a blob via `window.URL.createObjectURL` in `downloadFile`  was necessary because we did not have `?download=true`  four years ago (was introduced in 2020 in https://github.com/ipfs/go-ipfs/pull/7677),  and we also displayed progress bar or something. 

[...]

PR #1952, opened 2022-01-10
feat: better handling of 'unresponsive' event

hacdias:
Replaces the deprecated event 'crashed' by 'render-process-gone' and actually use the information correctly. In addition, I removed the `event.toString()` from the 'unresponsive' event because it does not contain additional info.

References:
- https://github.com/electron/electron/pull/23096
- https://www.electronjs.org/docs/latest/api/web-contents#event-unresponsive
- https://www.electronjs.org/docs/latest/api/web-contents#event-render-process-gone

Noticed this because on [#1951](https://github.com/ipfs/ipfs-webui/issues/1887), the logs just say `[object Object]`...

License: MIT

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -45,6 +45,8 @@ const createWindow = () => {
   })
 
   window.webContents.on('unresponsive', async () => {
+    logger.error('[web ui] the webui became unresponsive')
+
     const opt = showDialog({
       title: i18n.t('unresponsiveWindowDialog.title'),
       message: i18n.t('unresponsiveWindowDialog.message'),
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -38,8 +38,8 @@ const createWindow = () => {
     window.webContents.openDevTools()
   }
 
-  window.webContents.on('crashed', event => {
-    logger.error(`[web ui] crashed: ${event.toString()}`)
+  window.webContents.on('render-process-gone', (_, { reason, exitCode }) => {
+    logger.error(`[web ui] crashed: ${reason}, code: ${exitCode}`)
   })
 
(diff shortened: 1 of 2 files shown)


src/webui/index.js
```diff
@@ -39,12 +39,12 @@ const createWindow = () => {
...
-  window.webContents.on('crashed', event => {
-    logger.error(`[web ui] crashed: ${event.toString()}`)
+  window.webContents.on('render-process-gone', (_, { reason, exitCode }) => {
+    logger.error(`[web ui] crashed: ${reason}, code: ${exitCode}`)
   })
 
-  window.webContents.on('unresponsive', event => {
-    logger.error(`[web ui] unresponsive: ${event.toString()}`)
+  window.webContents.on('unresponsive', () => {
+    logger.error('[web ui] the webui became unresponsive')
```
lidel, line 47: Would it be possible to add a dialog here, where user can decide if they want to kill a rogue webui  process before it slows everything down to a crawl for some reason?

Found [a poc here](https://www.electronjs.org/docs/latest/api/web-contents#contentsforcefullycrashrenderer) (just needs i18n strings):

```
contents.on('unresponsive', async () => {
  const { response } = await dialog.showMessageBox({
    message: 'App X has become unresponsive',
    title: 'Do you want to try forcefully reloading the app?',
    buttons: ['OK', 'Cancel'],
    cancelId: 1
  })
  if (response === 0) {
    contents.forcefullyCrashRenderer()
    contents.reload()
  }
})
```

lidel (changes requested):
Thanks, LGTM, but small ask inline that improves UX

lidel (approved):
:ok_hand:, thank you!

PR #1955, opened 2022-01-14
refactor: remove checkCorsConfig 

hacdias:
This PR removes the 2 years old `checkCorsConfig` function that would magically remove `*` and `webui://-` from `Access-Control-Allow-Origin`. Defaults have changed and people had plenty of time to upgrade their IPFS Desktop version, so removing this old code shouldn't be an issue.

PR #1956, opened 2022-01-14
feat: add known webui addresses to CORS conf

hacdias:
This PR is built on top of #1955 for easiness.

I used the already existing `migrateConfig` function and built on top of it. I added revision checks so we apply the migration only once.

Closes ipfs/ipfs-webui#1880


hacdias:
Ping @lidel

hacdias:
Ping @lidel

lidel (approved):
Thanks for waiting.
Eyeballed this, and it should be safe to ship (go-ipfs special-cases Electron origins these days).

PR #1959, opened 2022-01-21
chore: smaller dep updates

lidel:
This bumps another set of  dependencies. (part of release prep)

## TODO

- [x] security patch of Electron 13.x
- [x] decrease trust in semver on NPM because _fool me 4 times_.. 
- [x] there is a bug when (1) `npm start` initializes a new repo (2) quit (3) `npm start` fails to start daemon correctly, webui is displayed in offline state
  - i confirmed this is a regression introduced by #1899  – perhaps ipfsd-ctl bug?
  - seems that the problem is that when ipfs-desktop quits it force-kills go-ipfs in a way that leaves `~/.ipfs/api` around. 
  - before #1899 `[daemon] removing api file` worked fine (after confirming that  multiaddr in `api` is from the local daemon who is offline, and not a remote node) – I suspect adding `ERR_CONNECTION_REFUSED` check introduced  a bug, probably invalid condition
  - yep that's it, fix in https://github.com/ipfs/ipfs-desktop/pull/1959/commits/1f4846ef87535ae36a3439f5e7265188e3dc454d


src/daemon/daemon.js
```diff
@@ -79,7 +79,7 @@ module.exports = async function (opts) {
     logger.info(`[daemon] PeerID is ${id}`)
     logger.info(`[daemon] Repo is at ${ipfsd.path}`)
   } catch (err) {
-    if (!err.message.includes('ECONNREFUSED') || !err.message.includes('ERR_CONNECTION_REFUSED')) {
+    if (!err.message.includes('ECONNREFUSED') && !err.message.includes('ERR_CONNECTION_REFUSED')) {
```
lidel, line 82: pixel-cat-salute

PR #1961, opened 2022-01-24
feat: electron 16

hacdias:


lidel:
We need tests before we switch, so let's continue in https://github.com/ipfs/ipfs-desktop/pull/1937

PR #1965, opened 2022-01-28
fix: electron version constant

hacdias:
Electron is a dev dependency. However, `devDependencies` is not available after packaging the app and it will throw an error right after starting IPFS Desktop.

The most fail proof way to obtain it is via `process.versions.electron`: https://www.electronjs.org/docs/latest/api/process#processversionselectron-readonly.

License: MIT


lidel (approved):
lgtm:

```
User-Agent: ipfs-desktop/0.18.1 (Electron 13.6.7)
```

PR #1966, opened 2022-01-28
feat: improved UX when opening IPFS URIs

### LINKED ISSUE #1646: Improve onboarding via OS-level protocol handler
Right now,  IPFS Desktop installs system-wide protocol handlers (dweb, ipfs, ipns) which converts `ipfs://<cid>` URI to gateway URL and opens that in default browser.

It is fine, but  may be confusing on the first use.

We could improve this behavior and show a dialog window, where user can choose how to open IPFS resource via radio-button-like selector:

- open from public gateway in the default web browser (default)
- open from a local gateway (show only if daemon in running)
- open in webui/Files (show only if daemon in running)
- open in webui/Explore (show only if daemon in running)
- (checkbox) remember the choice for the future
  - should be accompanied by a toggle in menubar menu: _"Ask how to handle IPFS URIs"_ so user can change their mind later

(thx @aschmahmann for noticing that onboarding is especially confusing if user's default browser is not the one with ipfs-companion)

hacdias:
This fixes #1646 by adding a dialog when opening an IPFS or IPNS with IPFS Desktop:

image

The UI matches the other dialogs we use on IPFS Desktop, which try to resemble a native dialog the most it can.

The option to ask when opening is also added under the preferences in the menubar tool:

image

Note: the Explore section does not support IPNS. Thus, I decided to open on the Files page if the user opens an IPNS URL and selects the Explore section by default. 

The other option is to ask separately for IPNS and IPFS. That is possible but will duplicate the options in the settings.


assets/locales/en.json
```diff
@@ -238,5 +240,14 @@
     "message": "Ongoing operation is taking more resources than expected. Do you wish to abort, and forcefully reload the interface?",
     "forceReload": "Yes, reload",
     "doNothing": "Do nothing"
+  },
+  "protocolHandlerDialog": {
+    "title": "Opening IPFS URIs",
```
lidel, line 245: I recently realized humans have no idea what URI is  :robot: 
```suggestion
    "title": "Opening IPFS address",
```

assets/locales/en.json
```diff
@@ -238,5 +240,14 @@
...
+    "message": "How would you like IPFS Desktop to open IPFS URIs?",
```
lidel, line 246:
```suggestion
    "message": "How would you like IPFS Desktop to open IPFS addresses?",
```

assets/locales/en.json
```diff
@@ -238,5 +240,14 @@
...
+    "browserPublicGateway": "In my default browser using a public gateway",
+    "browserLocalGateway": "In my default browser using a local gateway",
+    "filesScreen": "In the Files screen",
+    "exploreScreen": "In the Explore screen",
+    "rememberThisChoice": "Remember this choice for all IPFS URIs"
```
lidel, line 251:
```suggestion
    "rememberThisChoice": "Remember this choice for all IPFS addresses"
```

assets/locales/en.json
```diff
@@ -212,6 +213,7 @@
     "appPreferences": "App Preferences",
     "launchOnStartup": "Launch at Login",
     "openWebUIAtLaunch": "Open Web UI at Launch",
+    "askWhenOpeningIpfsURIs": "Ask When Opening IPFS URIs",
```
lidel, line 216:
```suggestion
    "askWhenOpeningIpfsURIs": "Ask How to Open IPFS Address",
```

lidel (changes requested):
Thank you, this is long time due! 

Small ask to replace "URI" with "address" (diffs inline)

> Note: the Explore section does not support IPNS. Thus, I decided to open on the Files page if the user opens an IPNS URL and selects the Explore section by default.
> The other option is to ask separately for IPNS and IPFS. That is possible but will duplicate the options in the settings.

Hm.. we don't want to duplicate settings, and opening Files instead of Explore looks like  a bug.
I think we should fix https://github.com/ipfs/ipfs-webui/issues/1884 before merging this PR, so there are no surprises around the way IPFS addresses are handled.

.. or just hide "Explore" option for now, so we don't need to wait for https://github.com/ipfs/ipfs-webui/issues/1884

hacdias:
@lidel 

> add IPFS cube logo to the Ask window 

Then, should we do the same for the "Download CID" window? We're using the same template after all.

> .. or just hide "Explore" option for now, so we don't need to wait for https://github.com/ipfs/ipfs-webui/issues/1884

I'll do that.

lidel (changes requested):
## Polishing the UX

I took it for a spin via  `npm start -- ipns://en.wikipedia-on-ipfs.org` to trigger the one-time choice dialog, current options are:

>  


My initial impressions:

- :art:  small UX asks
  - :ice_cube:  needs IPFS cube as a visual hint to indicate where the question comes from
     - > Then, should we do the same for the "Download CID" window? We're using the same template, after all.
 
       @hacdias  yes, sounds sensible and it would look good there too
  - :id:  could we show the UR Ithat triggered the dialog  as part of the question? I'm thinking  `How would you like to open IPFS addresses like  ipns://en.wikipedia-on-ipfs.org?`
    -  makes it easier to connect the dots ("i clicked on this ipns:// thing and this happened")
- :thought_balloon: I started questioning myself: is there a value in having separate browser options for public and local gateway? 
  - :anger: Beginner/casual users would not even know what this choice means
    - :bulb: this could be mitigated by adding "(safe default choice)" next to pre-selected "public gateway" option
  - :anger: If user selected "public"  but their browser has redirect to local (Brave, Companion) that looks like a bug, because they wanted public link, and ended up on local
  - :anger: If user selected "local" but the node is down, then links either fail or we need to use public gateway anyway 
      - I've seen folks strugging with this in ipfs-companion – the bug reports go away only  if we redirect to public gw when local is dead.   
  - :bulb: My gut feeling is that we should only have the public gateway option (simple "In my web browser") – kicking the routing decision to the browser. 
    - :green_heart: Browsers with IPFS support (Brave or IPFS Companion) would redirect public gateway link to local gateway anyway
    - :green_heart:  regular browsers would load the original URL, there is no footgun when local node is offline

Thoughts?

cc'ing @autonome @SgtPooki (not urgent, but before we ship this, would appreciate more :eyes: on this, both from the perspective of a native speaker, and general UX)

src/i18n.js
```diff
@@ -33,3 +33,7 @@ module.exports = async function () {
     ipcMain.emit('languageUpdated', lang)
   })
 }
+
+const promise = setupI18N()
+
+module.exports = async () => promise
```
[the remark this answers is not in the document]
hacdias: Oh, lol. Not really. I'll just export the promise itself.

hacdias:
> 🧊 needs IPFS cube as a visual hint to indicate where the question comes from

Done.

> could we show the UR that triggered the dialog as part of the question

This is easy to do, but we have to make sure it doesn't overflow or wrap to the next line, otherwise some window text will be hidden. The maximum size would also vary with translations so I'm not sure what's the best way to proceed here.

---

I will defer the remaining updates once we have @autonome's and @SgtPooki's feedback regarding the gateway options.

src/protocol-handlers.js
```diff
@@ -1,34 +1,138 @@
...
   } else if (url.startsWith('dweb:/ipns/')) {
-    return openLink('ipns', url.slice(11), base)
+    protocol = 'ipns'
+    part = url.slice(11)
+  } else {
+    return false
+  }
+
+  const action = await getAction()
+  let base = 'https://dweb.link'
```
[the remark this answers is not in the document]
hacdias: @lidel that is reasonably simple **if** we wait for the Web UI window to be ready. We have to wait for it to be ready to have that working. So if you open an IPFS link, IPFS Desktop has to boot up and wait for the Web UI window to be ready until the link is actually open. That could feel quite slow. That's why I thought about keeping a copy on Desktop's settings such that we do not need to wait.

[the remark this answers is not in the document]
hacdias: On [90b25e0](https://github.com/ipfs/ipfs-desktop/pull/1966/commits/90b25e083ba73fc7caebd6ad4bc9f5145ac7aa8a) I made a best effort attempt. If the web ui window hasn't been created yet, we use the default public gateway. I can, however, rearrange Desktop's startup logic to ensure the window always exists.

SgtPooki:
I agree 100% with all the callouts from lidel. A few additional comments:

> If user selected "public" but their browser has redirect to local (Brave, Companion) that looks like a bug, because they wanted public link, and ended up on local
> If user selected "local" but the node is down, then links either fail or we need to use public gateway anyway

I would reduce the number of options to only either: open in browser, open in ipfs desktop, and then let ipfs-companion/browser handle the rest, since they will override whichever decision is made in ipfs-desktop anyway. What do you think @lidel? And if we did do this, should the browser option always use local?

If a user is running ipfs-desktop, then they *should* have a local node running at the time ipfs-desktop opens the link in the browser, right?

lidel:
> I would reduce the number of options to only either: open in browser, open in ipfs desktop, and then let ipfs-companion/browser handle the rest, since they will override whichever decision is made in ipfs-desktop anyway. What do you think @lidel?

Yep, this sounds like the best way to do this.

> And if we did do this, should the browser option always use local?
> If a user is running ipfs-desktop, then they should have a local node running at the time ipfs-desktop opens the link in the browser, right?

If we have a generic "Open in browser"  then yes, 
ipfs-desktop could make a quick test to confirm local gateway is still online, and use its address instead of public one.
ipfs-webui is already doing that type of check (it checks if local gateway port is functional, it also confirms if OS supports subdomains on `localhost`, and sets gateway links to the best option available, falling back to public gateway URL as the last resort)

hacdias:
I will work on those changes then.

hacdias:
Ping @lidel

SgtPooki:
> I tested this on macOS. It seems that the localhost domain works when using Chrome but fails when using fetch (also seems to not work on Safari). Thus, even though Chrome may be the default browser, it won't use the localhost version.

@hacdias can you explain a little bit more about this? Are you saying chrome loads a URL with a localhost domain, but cannot make successful fetch calls on the page that is loaded?

Also, I will try to download the generated ipfs build on windows right now and update this comment.

src/protocol-handlers.js
```diff
@@ -1,31 +1,215 @@
-const { app, shell } = require('electron')
+const { app, shell, ipcMain } = require('electron')
```
SgtPooki, line 1: Many of  these changes seem unrelated to the general setting up of protocol-handlers. Is there a better place for a majority of the logic required to set up the IPFS protocol?

Maybe move `src/protocol-handlers.js` to `src/protocol-handlers/index.js` and then the logic for the ipfs protocol handling inside of `src/protocol-handlers/ipfs.js` ?

src/tray.js
```diff
@@ -16,14 +16,16 @@ const { CONFIG_KEY: PUBSUB_KEY } = require('./enable-pubsub')
 const { CONFIG_KEY: NAMESYS_PUBSUB_KEY } = require('./enable-namesys-pubsub')
 const { CONFIG_KEY: AUTO_GC_KEY } = require('./automatic-gc')
 const { CONFIG_KEY: AUTO_LAUNCH_WEBUI_KEY } = require('./webui')
+const { CONFIG_KEY: ASK_OPENING_IPFS_URIS } = require('./protocol-handlers')
```
SgtPooki, line 19: it looks like you're following convention, but config keys should all be defined in the same place, can you move us towards better practices by putting your config key into `src/common/configKeys.js` ?

SgtPooki (changes requested):
Thanks for the updates @hacdias . Many comments are questions because I'm newer to this repo, but the ctx refactor has helped me to gain a bit of understanding of the current problems we face with the ipfs-desktop, so I hope the non-question comments are helpful =)

src/tray.js
```diff
@@ -16,14 +16,16 @@ const { CONFIG_KEY: PUBSUB_KEY } = require('./enable-pubsub')
...
 
 const CONFIG_KEYS = [
   AUTO_LAUNCH_KEY,
   AUTO_LAUNCH_WEBUI_KEY,
   AUTO_GC_KEY,
   SCREENSHOT_KEY,
   PUBSUB_KEY,
-  NAMESYS_PUBSUB_KEY
+  NAMESYS_PUBSUB_KEY,
+  ASK_OPENING_IPFS_URIS
```
SgtPooki, line 28: @lidel @achingbrain This is why we need to use trailing commas. How can I go about updating the styling across our repos so that we can allow/require trailing commas in our code? These diffs don't need to be this noisy. git blaming is also a PITA when we try to find out who originally added NAMESYS_PUBSUB_KEY, or whatever the non-trailing-comma-having previous line ends up being in PRs.

assets/locales/en.json
```diff
@@ -33,6 +33,7 @@
   "yes": "Yes",
   "no": "No",
   "close": "Close",
+  "continue": "Continue",
```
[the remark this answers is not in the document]
hacdias: No, when we update Transifex, those are added automatically.

src/dialogs/prompt/index.js
```diff
@@ -27,6 +65,7 @@ module.exports = async function showPrompt (options) {
     show: false,
     width: 350,
     height: 330,
+    useContentSize: true,
```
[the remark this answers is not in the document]
hacdias: Yes. `useContentSize`, as the linked docs says, means:

> The width and height would be used as web page's size, which means the actual window's size will include window frame's size and be slightly larger

`useContentSize` is used in order to ensure that the content part of the window is actually the height we set. Otherwise, the height we set will also include the frame and menu size's, which we can't estimate because they vary cross-platform.

src/dialogs/prompt/styles.js
```diff
@@ -60,9 +60,27 @@ input {
...
+  margin-right: 0.25rem;
+}
 #buttons {
   text-align: right;
 }
+#logo {
+  width: 4rem;
+  margin: 0 auto .5rem;
+  display: block;
+}
```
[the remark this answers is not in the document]
hacdias: These are the styles for prompts only. This is completely unrelated from Web UI. This is the **only** other window, besides Web UI, that we open that uses custom (HTML+CSS+JS) things. Other dialogs are native dialogs.

> we need to worry about the above selectors leaking styles globally? If not, would you mind clarifying why not?

This variable is only loaded on the prompt. I don't understand what "leaking globally" would mean in this context as the prompt window is isolated from any other windows.

src/dialogs/prompt/template.js
```diff
@@ -1,14 +1,23 @@
 const { styles } = require('./styles')
 
-module.exports = ({ message, defaultValue, buttons, id }) => (`<!DOCTYPE html>
+const fs = require('fs-extra')
+const path = require('path')
+
+const ipfsLogoPath = path.join(__dirname, '../../../assets/icons/ipfs.svg')
+const ipfsLogo = 'data:image/svg+xml;base64,' + fs.readFileSync(ipfsLogoPath).toString('base64')
```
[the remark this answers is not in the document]
hacdias: > these should probably be in commons/consts, as this logo will probably need to be used elsewhere.

Sure.

> Also, did we really not already have a logo available? that's odd.

We did, but not a PNG of this size/dimensions. The other logos available in the repository under `assets/` are tray icons with minuscule sizes.

src/dialogs/prompt/index.js
```diff
@@ -5,14 +5,52 @@ const dock = require('../../utils/dock')
...
+        str += '</div>'
+
+        break
+      case 'text':
+        str += `<input type="text" name="${name}" id="${name}" value="${defaultValue}" />`
+    }
+
+    str += '</div>'
+    return str
+  }).join('\n')
```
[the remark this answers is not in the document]
hacdias: > Is this the idiomatic way we handle HTML inside of the ipfs-desktop project?

I'm open for suggestions. I think this directory (`prompt`) is the only place where we use HTML to render the prompt. Unless we start using a rendering engine here, I don't know what else we could do.

src/i18n.js
```diff
@@ -33,3 +33,5 @@ module.exports = async function () {
     ipcMain.emit('languageUpdated', lang)
   })
 }
+
+module.exports = setupI18n()
```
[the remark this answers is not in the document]
hacdias: The problem here is that this dialog _might_ be the first action IPFS Desktop will take. When you open an `ipfs://` URL on your computer and IPFS Desktop is shut down, it will start up and show this dialog. If the i18n isn't loaded, it won't work. However, I saw your PR and yes, there are better ways to do it. Perhaps we can merge #2114 before and make sure to rebase this correctly to ensure the i18n stuff is loaded before any window is open.

src/index.js
```diff
@@ -11,8 +11,8 @@ if (process.env.NODE_ENV === 'test') {
 const fixPath = require('fix-path')
```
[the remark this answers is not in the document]
hacdias: Yes!

src/tray.js
hacdias: We can do that, would be great. However, outside of this PR, which is already a bit big.

src/dialogs/prompt/styles.js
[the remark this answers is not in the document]
SgtPooki: makes sense. thanks!

src/i18n.js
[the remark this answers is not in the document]
SgtPooki: I hope to get the ctx refactor done late today if my ESM migration of js-pinning-service-http-client can finish today.

I do want to clarify that even if the dialog is required as the first time IPFS desktop is opened, we can keep the prior function as is, but cache the result (when appropriate) and await any pre-reqs inside the place where we need it; not globally. In general, we should

1. Make everything async. The refactor ctx PR migrates many things to async, but I'm not as comfortable with the app as I should be so I tried not to change things unnecessarily.
2. Execute on first use, not on require/import (whether in src/index or src/dialogs/prompt/index), caching the resulting value if re-evaluations are not necessary. `if (result != null) return result` at the top of everything.
3. Await each required dependency & pre-req within each module. 

With these three simple rules, without even trying, anything/everything we need is 99% of the way to being [SOLID](https://en.wikipedia.org/wiki/SOLID). This is essentially the same as dynamic imports, but in the form of asynchronously awaiting required dependencies & pre-req processes.


https://github.com/ipfs/ipfs-desktop/pull/1966#discussion_r866967386 mentions more about doing this.

src/tray.js
SgtPooki: sounds fair!

hacdias:
@SgtPooki I think I addressed all of your comments!

src/dialogs/prompt/html.js
```diff
@@ -0,0 +1,122 @@
...
+        str += `<input type="text" name="${name}" id="${name}" value="${defaultValue}" />`
+    }
+
+    str += '</div>'
+    return str
+  }).join('\n')
+}
+
+/**
+ * Generates a base64 encoded URI with the HTML content of a prompt window.
```
SgtPooki, line 70: Why do we need to do this as a base64 encoded URI?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+        type: 'radio',
+        name: 'action',
+        defaultValue: DEFAULT_ACTION,
+        labels: {
+          [ACTION_OPTIONS.OPEN_IN_BROWSER]: i18n.t('protocolHandlerDialog.openInBrowser'),
+          [ACTION_OPTIONS.OPEN_IN_IPFS_DESKTOP]: i18n.t('protocolHandlerDialog.openInIpfsDesktop')
+          // [ACTION_OPTIONS.BROWSER_PUBLIC_GATEWAY]: i18n.t('protocolHandlerDialog.browserPublicGateway'),
+          // [ACTION_OPTIONS.BROWSER_LOCAL_GATEWAY]: i18n.t('protocolHandlerDialog.browserLocalGateway'),
+          // [ACTION_OPTIONS.FILES_SCREEN]: i18n.t('protocolHandlerDialog.filesScreen')
+          // [ACTION_OPTIONS.EXPLORE_SCREEN]: i18n.t('protocolHandlerDialog.exploreScreen')
```
SgtPooki, line 43: remove?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+const CONFIG_KEY = 'askWhenOpeningIpfsURIs'
+const CONFIG_KEY_ACTION = 'openIpfsURIsAction'
+
+const ACTION_OPTIONS = {
+  OPEN_IN_BROWSER: 'openInBrowser',
+  OPEN_IN_IPFS_DESKTOP: 'openInIpfsDesktop'
+  // BROWSER_PUBLIC_GATEWAY: 'browserPublicGateway',
+  // BROWSER_LOCAL_GATEWAY: 'browserLocalGateway',
+  // FILES_SCREEN: 'filesScreen'
+  // EXPLORE_SCREEN: 'exploreScreen'
```
SgtPooki, line 18: remove?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+}
+
+const DEFAULT_ACTION = ACTION_OPTIONS.OPEN_IN_BROWSER
+
+async function getAction () {
+  const ask = store.get(CONFIG_KEY, true)
+  if (!ask) {
+    return store.get(CONFIG_KEY_ACTION, DEFAULT_ACTION)
+  }
```
SgtPooki, line 27: I assume this is checking if we have a saved config preference, and if not, we're then returning the default action?

Does that mean that if we *do* have a config preference, that the below prompt shows? What am I missing?

Can we change `ask` to a more explicit name that explains what it's doing?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+    ],
+    window: {
+      width: 500,
+      height: 218
+    }
+  })
+
+  if (button !== 0) {
+    return
+  }
```
SgtPooki, line 65: what does this mean? if button is not zero, we return? I'm assuming zero here means "Ignore" or "cancel" ?

Can you add a comment for noobs like me? :)

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+
+  // Handle if the app started running now, and a link
+  // was sent to be handled.
+  argvHandler(process.argv, ctx)
+
+  // Handle URLs in macOS
+  app.on('open-url', (event, url) => {
+    event.preventDefault()
+    parseUrl(url, ctx)
+  })
```
SgtPooki, line 133: do we need an `IS_MAC` check here?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+      width: 500,
+      height: 218
+    }
+  })
+
+  if (button !== 0) {
+    return
+  }
+
+  const action = input.action || DEFAULT_ACTION
```
SgtPooki, line 67: nit: `const {remember, action} = input`

Also, do we need the `|| DEFAULT_ACTION` here if we have `defaultValue: DEFAULT_ACTION` above?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+    return true
+  }
+
+  if (action === ACTION_OPTIONS.OPEN_IN_IPFS_DESKTOP) {
+    ctx.launchWebUI(`/${parsed.protocol}/${parsed.hostname}${parsed.path}`, { focus: true })
+    return true
+  }
+
+  return false
+}
```
SgtPooki, line 98: A few questions:

1. What do we do if there is an error in one of the awaited functions?
2. What does return true/false mean?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+  let handled = false
+
+  for (const arg of argv) {
+    if (await handleUrl(arg, ctx)) {
+      handled = true
+    }
+  }
+
+  return handled
+}
```
SgtPooki, line 110: How does the return value of true/false change the meaning of what's happened? Do we need a return value?

src/protocol-handlers/index.js
```diff
@@ -0,0 +1,138 @@
...
+  if (store.get(CONFIG_KEY, null) === null) {
+    store.set(CONFIG_KEY, true)
+    ipcMain.emit('configUpdated')
+  }
+
+  createToggler(CONFIG_KEY, () => true)
+
+  // Handle if the app started running now, and a link
+  // was sent to be handled.
+  argvHandler(process.argv, ctx)
```
SgtPooki, line 127: we're not using the return value of argvHandler. Is that intentional? Do we need a return value?

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+ * @property {string} hostname
+ * @property {string} path
+ */
+
+/**
+ * Parses an IPFS/IPNS/dWeb URL to be handled by IPFS Desktop.
+ *
+ * @param {string} url
+ * @returns {ParsedUrl}
+ */
```
SgtPooki, line 21: I see a `return null` which means `parseUrl` does not only return a `ParsedUrl` type. Be sure to include null in the possible returned values.

I believe a simple `// @ts-check` pragma at the top of the page would display this problem if your editor is configured properly.

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+    return null
+  }
+
+  if (hostname.includes('/')) {
+    const [first, ...rest] = hostname.split('/')
+    hostname = first
+    path = '/' + rest.join('/')
+  }
+
+  return { protocol, hostname, path }
```
SgtPooki, line 49: using the `URL` class should help here.

We should add some tests around this method to ensure we're parsing things properly.

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+async function getPublicGatewayUrl (ctx) {
+  if (!ctx.webui) {
+    // Best effort. If the Web UI window wasn't created yet, we just return the default
+    // gateway.
+    return DEFAULT_GATEWAY
+  }
+
+  return await ctx.webui
+    .webContents
+    .executeJavaScript('JSON.parse(localStorage.getItem("ipfsPublicGateway")) || "https://dweb.link"', true)
```
SgtPooki, line 61: instead of the OR in the executed javascript, can we do the OR with something like

```js

/** @type {string} */
let publicGatewayUrl
try {
   publicGatewayUrl = await ctx.webui.webContents.executeJavaScript('JSON.parse(localStorage.getItem("ipfsPublicGateway"))', true))
} catch {
  publicGatewayUrl = 'https://dweb.link'
}

return publicGatewayUrl
```

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+
+  return toUri(gateway)
+}
+
+const checkIfGatewayUrlIsAccessible = async (url) => {
+  try {
+    const { status } = await fetch(
+    `${url}/ipfs/bafkqae2xmvwgg33nmuqhi3zajfiemuzahiwss`
+    )
+    return status === 200
```
SgtPooki, line 87: do we need a status=200 or could we check for `ok`?

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+
+// Separate test is necessary to see if subdomain mode is possible,
+// because some browser+OS combinations won't resolve them:
+// https://github.com/ipfs/go-ipfs/issues/7527
+const checkIfSubdomainGatewayUrlIsAccessible = async (url) => {
+  try {
+    url = new URL(url)
+    url.hostname = `bafkqae2xmvwgg33nmuqhi3zajfiemuzahiwss.ipfs.${url.hostname}`
+    const { status } = await fetch(url.toString())
+    return status === 200
```
SgtPooki, line 101: would using `ok` be sufficient or do we require an explicit 200?

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+// Separate test is necessary to see if subdomain mode is possible,
+// because some browser+OS combinations won't resolve them:
+// https://github.com/ipfs/go-ipfs/issues/7527
+const checkIfSubdomainGatewayUrlIsAccessible = async (url) => {
+  try {
+    url = new URL(url)
+    url.hostname = `bafkqae2xmvwgg33nmuqhi3zajfiemuzahiwss.ipfs.${url.hostname}`
```
SgtPooki, line 99: What is the hardcoded CID in `bafkqae2xmvwgg33nmuqhi3zajfiemuzahiwss.ipfs.${url.hostname}`?

Should that be saved in consts.js?

SgtPooki (changes requested):

src/dialogs/prompt/html.js
hacdias: We don't need, but it seems to be the easier solution. The other solution is to start a server with an HTML file and somehow costumize the page on the fly. This seems easier for a simple prompt.

src/protocol-handlers/index.js
hacdias: This checks:
1. Am I supposed to ask when opening IPFS URIs? Default is yes.
2. Then, if you are not supposed to ask, it returns either the default action, or the user's preference.
Renamed to `askWhenOpeningUri`. Feel free to suggest something else.

src/protocol-handlers/index.js
hacdias: No. The event is only triggered on macOS: https://www.electronjs.org/docs/latest/api/app#event-open-url-macos

src/protocol-handlers/urls.js
hacdias: Code borrowed from Web UI: https://github.com/ipfs/ipfs-webui/blob/4bf78a47fcb38872edbe55a5046ae3c7d3fe1a10/src/bundles/config.js#L113-L138

Not sure if we need explicit 200 or any OK suffices. /cc @lidel

src/protocol-handlers/urls.js
hacdias: Same as before: I did not write this code, but it seems that it is some safe CID (https://github.com/ipfs/ipfs-webui/commit/ebda7644dbf1a4cf0712b31a5fa3a68ccf3aa470) for "Welcome to IPFS :-)"

src/protocol-handlers/urls.js
hacdias: https://github.com/ipfs/ipfs-desktop/pull/1966#discussion_r922080264

src/protocol-handlers/index.js
hacdias: 1. The error propagates up and you'll get a nice prompt saying there was an error.
2. true/false means whether or not the URL was handled or not.

src/protocol-handlers/index.js
hacdias: Yes. It is used in other files.

https://github.com/ipfs/ipfs-desktop/blob/a861ffb8e7b3d0993690aaed15f15c6af74263ed/src/second-instance.js#L1-L17

src/protocol-handlers/index.js
hacdias: Yes. See above.

src/protocol-handlers/urls.js
hacdias: Try catch won't work here. Do you think this would be an improvement:

```javascript
/** @type {string|null} */
  let publicGatewayUrl = await ctx.webui.webContents.executeJavaScript('JSON.parse(localStorage.getItem("ipfsPublicGateway"))')
  if (!publicGatewayUrl) {
    publicGatewayUrl = 'https://dweb.link'
  }
```
?

src/protocol-handlers/urls.js
hacdias: I think I used (or tried) before. But I got something along these lines, which seems much more complicated than the current approach:

```js
function parseUrl (urlStr) {
  let protocol = null
  let hostname = null
  let path = '/'

  try {
    const url = new URL(urlStr)

    if (['ipfs:', 'ipns:'].includes(url.protocol)) {
      protocol = url.protocol.slice(4)
      hostname = url.pathname
    } else if (url.protocol === 'dweb:') {
      if (url.pathname.startsWith('/ipfs/')) {
        protocol = 'ipfs'
        hostname = url.pathname.slice(6)
      } else if (url.pathname.startsWith('/ipns/')) {
        protocol = 'ipns'
        hostname = url.pathname.slice(6)
      } else {
        return null
      }
    } else {
      return null
    }
  } catch (_) {
    return null
  }

  if (hostname.startsWith('/')) {
    hostname = hostname.substring(1)
  }

  if (hostname.includes('/')) {
    const [first, ...rest] = hostname.split('/')
    hostname = first
    path = '/' + rest.join('/')
  }

  return { protocol, hostname, path }
}
```

src/protocol-handlers/urls.js
SgtPooki: I think that works, along with wrapping *any and all* JSON.parse, and awaits, in try/catch blocks.

hacdias:
@SgtPooki solved the conflicts. It seems that the CI is taking way too long downloading Web UI for the tests. We may have to restart them later.

src/protocol-handlers/urls.js
lidel: This is a CID which has `Welcome to IPFS :-)` string inlined inside  `identity`  Multihash.
Main reason for using this exotic multihash type is to avoid hitting DHT/datastore – gateway can respond immediatelly after decoding CID.

src/protocol-handlers/urls.js
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Get the gateway URL. Logic borrowed from Web UI.
+ * Please check https://github.com/ipfs/ipfs-webui/pull/1591.
+ *
+ * @param {object} ctx
+ * @param {ParsedUrl} parsedUrl
+ * @returns
+ */
+async function getGatewayUrl (ctx, { protocol, hostname, path }) {
```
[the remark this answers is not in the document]
lidel: Afaik no. 
There are things like https://github.com/ipfs/js-ipfs-utils (legacy, does not seem to be maintained) and  https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-core-utils/src (part of js-ipfs), but they pull-in too many dependencies.

lidel (changes requested):
@SgtPooki @hacdias I just realized this PR comes with an awful UX  footgun :crying_cat_face: 


If user tries to open `ipns://en.wikipedia-on-ipfs.org` and decides to do that in IPFS WebUI, it will fail because ipfs-webui is not optimized for showing huge HAMT-sharded UnixFS directories, and `/ipns/en.wikipedia-on-ipfs.org/wiki/` is one (has ~20M items). 


Due to this, I feel we need to mark this as blocked until we add a smart HAMT handling.

lidel:
This ended up no longer feasible, see https://github.com/ipfs/ipfs-desktop/issues/1646#issuecomment-1796859438

PR #1976, opened 2022-02-09
fix: use new glob source API

### LINKED ISSUE #1973: MS Windows: Error: Pattern must be a string
0.19.0 RC2 @ 3bc3a8a29e16473a11786fa0b9f1d42365f436af fail to import a single file via context menu in Windows' File Explorer with error ` Error: Pattern must be a string` in error.log

Error comes from add-to-ipfs.js (probably similar problem as one fixed in https://github.com/ipfs/ipfs-desktop/commit/13de17403fcd1892a0830b3da0199d116a14d899)

hacdias:
Fixes #1973 and adds unit tests. I also stopped using `object.addLink` and used the same strategy as in https://github.com/ipfs/ipfs-webui/pull/1890.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "is-ipfs": "6.0.2",
         "it-all": "^1.0.6",
         "it-concat": "^2.0.0",
+        "it-last": "^1.0.6",
         "multiaddr": "10.0.1",
         "multiaddr-to-uri": "8.0.0",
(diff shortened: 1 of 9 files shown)


hacdias:
@lidel this should fix #1973.

lidel (approved):
Thank you, does the trick on both Windows and macOS :+1:

PR #1979, opened 2022-02-15
fix: macos autoupdate flow

lidel:
This PR fixes auto-update flow on macOS, which seems to have a regression in v0.19.1.

- It removes macOS-specific code paths and replaces them with one unified logic for all three platforms (Windows/macOS/AppImage).
- Improves the flow by giving user option to dismiss upgrade if they are in the middle of something, and if they don't restart, they will be prompted again at a later time
- It also fixes a bug where tray disappears due to GC (it only worked because we had reference in ctx, but it could break at any time)


lidel:
- new flow tested in https://github.com/filecoin-project/filecoin-station/releases/tag/v0.0.4 – works as expected on macOS and  Windows
- merging to fix macOS asap.

lidel:
Confirmed this PR fixed autoupdate: [v0.19.2](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.19.2) updates to [v0.19.3](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.19.3) as expected, and if update is ignored, it gets applied on exit.

extraunlimited:
since the update I cannot connect to the IPFS API on macOS anymore!

PR #1982, opened 2022-02-17
feat: show repo migration (go-ipfs 0.12.0)

hacdias:
Closes #1940 and closes #1823

Screenshot 2022-02-25 at 17 51 21

Screenshot 2022-02-25 at 17 50 51

PR #1997, opened 2022-02-28
feat: go-ipfs 0.12.0

guseggert:


hacdias:
Hi @guseggert! This is already included in #1982, which is pending review.

PR #2011, opened 2022-03-04
fix: gui error report title

lidel:
This PR makes issues created by users from the error dialog more useful  by setting the title to the very first like in the stack trace, which usually includes a useful error message.


lidel:
@SgtPooki  FYSA this is what you suggested during triage, should  make GUI triage easier going forward.
I will ship this with Monday's release.

PR #2047, opened 2022-03-18
chore: remove unused and update dependencies

hacdias:
I found a bunch of rogue dependencies that are not being used _anywhere_. My best guess is that they were used at some point. Some of them are deprecated and haven't been touched for years. I checked them all and couldn't find any usage in the current codebase.


package.json
```diff
@@ -82,10 +78,8 @@
     "fs-extra": "^10.0.1",
     "go-ipfs": "0.12.0",
     "i18next": "^21.6.13",
-    "i18next-electron-language-detector": "0.0.10",
     "i18next-fs-backend": "1.1.4",
     "i18next-icu": "^2.0.3",
-    "intl-messageformat": "^9.11.4",
```
lidel, line 88: This one needs to stay, it is a peer dependency of https://github.com/i18next/i18next-icu#introduction

lidel (changes requested):

lidel (approved):
Thanks!

pngegg

PR #2055, opened 2022-03-21
fix: lowpower config in Swarm.ConnMgr

### LINKED ISSUE #2039: ship with low power profile as default to reduce CPU usage

I consistently see IPFS using ~5-11% CPU while I'm not interacting with it.

Ship w/ low power profile as default.

**Describe alternatives you've considered**

Not running IPFS typically reduces the CPU usage to zero.

**Additional context**

Apparently humans are *not* 80% water. It's a myth.

lidel:
Closes #2039 by applying lowpower values to `Swarm.ConnMgr` for new and existing users:

```
  config.Swarm.ConnMgr.GracePeriod = '1m'
  config.Swarm.ConnMgr.LowWater = 20
  config.Swarm.ConnMgr.HighWater = 40
```

I've run it for a while and works fine, peer count jumps to few hundreds when needed, and then dials down after a minute or two. 

This is a pretty safe one-time migration.
We can always adjust it further in the future releases.

### DIFF
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -46,9 +46,9 @@ function applyDefaults (ipfsd) {
   config.Swarm = config.Swarm || {}
   config.Swarm.DisableNatPortMap = false
   config.Swarm.ConnMgr = config.Swarm.ConnMgr || {}
-  config.Swarm.ConnMgr.GracePeriod = '300s'
-  config.Swarm.ConnMgr.LowWater = 50
-  config.Swarm.ConnMgr.HighWater = 300
+  config.Swarm.ConnMgr.GracePeriod = '1m'
+  config.Swarm.ConnMgr.LowWater = 20
+  config.Swarm.ConnMgr.HighWater = 40
 
   config.Discovery = config.Discovery || {}
   config.Discovery.MDNS = config.Discovery.MDNS || {}
@@ -76,7 +76,7 @@ function getHttpPort (addrs) {
 // This is the place where we execute fixes and performance tweaks for existing users.
 function migrateConfig (ipfsd) {
   // Bump revision number when new migration rule is added
-  const REVISION = 3
+  const REVISION = 4
   const REVISION_KEY = 'daemonConfigRevision'
   const CURRENT_REVISION = store.get(REVISION_KEY, 0)
 
@@ -131,7 +131,22 @@ function migrateConfig (ipfsd) {
     }
   }
 
-  // TODO: update config.Swarm.ConnMgr.*
+  if (CURRENT_REVISION < 4) {
+    // lower ConnMgr https://github.com/ipfs/ipfs-desktop/issues/2039
+    const { GracePeriod, LowWater, HighWater } = config.Swarm.ConnMgr
+    if (GracePeriod === '300s') {
+      config.Swarm.ConnMgr.GracePeriod = '1m'
+      changed = true
+    }
+    if (LowWater > 20) {
+      config.Swarm.ConnMgr.LowWater = 20
+      changed = true
+    }
+    if (HighWater > 40) {
+      config.Swarm.ConnMgr.HighWater = 40
+      changed = true
+    }
+  }
 
   if (changed) {
     try {
--- a/test/e2e/launch.e2e.test.js
+++ b/test/e2e/launch.e2e.test.js
@@ -170,6 +170,27 @@ test.describe.serial('Application launch', async () => {
     ])
   })
 
+  test('applies config migration (ConnMgr)', async () => {
+    // create preexisting, initialized repo and config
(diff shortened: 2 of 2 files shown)


RubenKelevra:
Agreed, I use similar values for a long time. But I would recommend to extend the grace period to something like 3 minutes to avoid that connections get dropped too early.

One minute seems a bit too restrictive if you consider usual user behaviour in a browser.

RubenKelevra:
@lidel have you seen my comment? :)

lidel:
@RubenKelevra yes, but we already increased it to 1m, when compared to go-ipfs defaults.
If you feel this should be increased, feel free to open a new PR with rationale why (the benefit is unclear to me, but i am happy to learn)  :pray:

RubenKelevra:
@lidel let's discuss this here. I'm happy to make a PR after we came to a decision :)

So the idea of setting it to 3 minutes is simple:

If you open a webpage etc. you open a lot of new connections to search in the network for the data you're looking for.

After this, a user usually "uses" a webpage for a while until the next navigation event. 

If the navigation event happens after one minute has passed, the chance is high that we have already terminated the connections, as they are so new that they are not deemed "useful" for IPFS (yet). If the navigation is within the same webpage we will significantly slow down the browsing experience, as bitswap can't flood-ask the node which provided the current webpage for the next one, but we instead need to go through the DHT to find the providing node.

This might have been an issue before, but probably not as serious, as the connections were not cut down to such a low amount. So there was a much larger chance that a connection stays open if we currently fetch data from them in short intervals.

lidel:
This reasoning is sensible, but the "3 minutes" value seems arbitrary, just like "1 minute" was, right? I could say that either the user navigates to the next page within 1 minute, or stays there for longer (e.g. spends 4 or 10 minutes reading an article or watching a video) :)

I am ok with changing `GracePeriod` but we should avoid gut feelings going forward, and base value on some UX research data. 

Namely: how long people spend on a web page withing a single Origin, on average/median? 

I think such papers/stats exist somewhere, if you have time to dig them out, I'm happy to accept change. :+1: 
(it would also be good basis for creating `desktop` profile in go-ipfs itself with these values)

RubenKelevra:
Well if the user streams a video the connection(s) will provide a lot more data. So if the website and the video is send from the same source there shouldn't be an issue, as IPFS would surely rate this node more useful than those used find the server via DHT query in the first place.

There are some (a bit older) statistics about the "length of stay" on a single webpage which is quoted to be 40 seconds.[0]

I picked 3 minutes because it "feels" sufficiently above the 40 seconds average and is close to the quoted read time I often see on news articles. I found some statistics on that.[1]

Additionally I looked up the default keep alive timeout for Firefox[2] which is like 2 minutes. Given that we need to do substantially more work if we disconnected I think it is safe to assume that 3 minutes is a good value. 



[0] https://de.m.wikipedia.org/wiki/Verweildauer_(Internet)
[1] https://www.google.com/amp/s/blog.chartbeat.com/2020/04/15/most-read-articles-engagement-data/amp/
[2] http://kb.mozillazine.org/Network.http.keep-alive.timeout

RubenKelevra:
@lidel wrote
> (it would also be good basis for creating `desktop` profile in go-ipfs itself with these values)

You're sure we need an additional profile? I always thought the default profile is for desktops (and that's the reason why we have a server profile).

lidel:
To be honest, I am not sold on changing this to 3m, given the average stay stat  being under one minute :see_no_evil: 

HTTP keep-alive does not apply here, we have N and not 1 connection to worry about. 

My main concern right now is around how the Grace Period works: AFAIK it will protect connection no matter if  it was used for bitswap or not all.  Raising this to 3m means the user will quickly reach thousands of peers and in theory,  kill their cheap router even faster.   Browsing Wikipedia and some other IPFS websites will produce over 1k of peers, and these won't get closed fast enough.

I propose we park this until:
-  go-ipfs 0.13 ships `Swarm.ResourceMgr` – that will allow us to set hard connection limits
- we are able to set separate Grace Period  for peers that sent us some data over bitswap 

> You're sure we need an additional profile? I always thought the default profile is for desktops (and that's the reason why we have a server profile).

Yes, we need it to remove ambiguity like this, and to make it easy to undo `server` settings.

lidel:
re: 3m proposal, see my reply in https://github.com/ipfs/ipfs-desktop/pull/2096#pullrequestreview-936752564

> You're sure we need an additional profile? I always thought the default profile is for desktops (and that's the reason why we have a server profile).

Yes, we need it to remove ambiguity like this, and to make it easy to undo `server` settings.

RubenKelevra:
Ah okay. So the default stays there as "works in most cases" while the server/desktop profiles are optimized for the usecase.

PR #2074, opened 2022-03-25
fix: return error instead of throw

### LINKED ISSUE #2069: [gui error report] Error: ports already being used
👉️ Please describe what you were doing when this error happened.

**Specifications**

- **OS**: win32
- **IPFS Desktop Version**: 0.20.3
- **Electron Version**: 17.1.0
- **Chrome Version**: 98.0.4758.102

**Error**

```
Error: ports already being used
    at checkPorts (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\config.js:302:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async module.exports (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\daemon.js:171:5)
    at async startIpfs (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\index.js:46:17)
    at async module.exports (C:\Program Files\IPFS Desktop\resources\app.asar\src\daemon\index.js:114:3)
    at async run (C:\Program Files\IPFS Desktop\resources\app.asar\src\index.js:83:5)
```

hacdias:
Fixes #2069

License: MIT

### DIFF
--- a/src/daemon/daemon.js
+++ b/src/daemon/daemon.js
@@ -168,7 +168,11 @@ async function startIpfsWithLogs (ipfsd) {
 module.exports = async function (opts) {
   const { ipfsd, isRemote } = await spawn(opts)
   if (!isRemote) {
-    await checkPorts(ipfsd)
+    try {
+      await checkPorts(ipfsd)
+    } catch (err) {
+      return { err }
+    }
   }
 
   let errLogs = await startIpfsWithLogs(ipfsd)


lidel (changes requested):
Could be me not understanding how to test this, but it does not seem to work as expected.

If I create `.ipfs/api` with `/ip4/127.0.0.1/tcp/5006` and then start `npx serve -l 5006` to simulate non-HTTP  service on that port, and then try to start ipfs-desktop, I see no error.

Electron window with ipfs-webui is displayed in broken state and  `[ipfsd] start daemon HTTPError:` is logged to the console, but no dialog is displayed to the user.

hacdias:
@lidel can you try on `main`? Because that also doesn't trigger an error on `main` for me.

hacdias:
Ping @lidel

lidel (approved):
For some reason, I was under the impression this PR introduced regression around API port, but if the behavior was the same in `main` then fine merge this. API issues are separate things, need separate issues/PRs  if they are a real problem (seems no?)

PR #2075, opened 2022-03-25
chore: electron 17.2 and related deps

lidel:
- [x]  electron*


lidel:
Continued in https://github.com/ipfs/ipfs-desktop/pull/2047

PR #2096, opened 2022-04-07
expand the protection interval for new connections to 3m

RubenKelevra:
The protection interval was previously set to 1m, which was deemed too 
low.

Rationale: 
https://github.com/ipfs/ipfs-desktop/pull/2055#issuecomment-1090327377


lidel (changes requested):
To be honest, I am not sold on changing this to 3m, given the average stay stat  being under one minute :see_no_evil: 

HTTP keep-alive does not apply here, we have N and not 1 connection to worry about. 

My main concern right now is around how the Grace Period works: AFAIK it will protect connection no matter if  it was used for bitswap or not all.  Raising this to 3m means the user will quickly reach thousands of peers and in theory,  kill their cheap router even faster.   Browsing Wikipedia and some other IPFS websites will produce over 1k of peers, and these won't get closed fast enough.

I propose we park this until:
-  go-ipfs 0.13 ships `Swarm.ResourceMgr` – that will allow us to set hard connection limits
- we are able to set separate Grace Period  for peers that sent us some data over bitswap

RubenKelevra:
@lidel yeah I understand your concerns, but I got one of those pesky devices as my router and don't have issue's since I did this.

Connections stay somewhere around 400 max for me. But since they are only established and then terminated after a while instead of replaced by other new ones if they terminate, the router seems to be fine with that.

In my case it's a 4G router from TP-Link.

RubenKelevra:
@lidel yeah I understand your concerns, but I got one of those pesky devices as my router and don't have issue's since I did this.

Connections stay somewhere around 400 max for me. But since they are only established and then terminated after a while instead of replaced by other new ones if they terminate, the router seems to be fine with that.

In my case it's a 4G router from TP-Link.


> Browsing Wikipedia and some other IPFS websites will produce over 1k of peers, and these won't get closed fast enough.

Let me check that.

RubenKelevra:
Yeah I agree, it creates a lot of connections while loading Wikipedia. Not sure what's the issue there, as it is also extremely slow, with loading times between 10 seconds and one minute.

So I would agree that 1 minute is probably the better value and something like 3 minutes for a "Grace Period for peers that sent us some data over bitswap". :)

src/daemon/config.js
```diff
@@ -47,7 +47,7 @@ function applyDefaults (ipfsd) {
   config.Swarm = config.Swarm || {}
   config.Swarm.DisableNatPortMap = false
   config.Swarm.ConnMgr = config.Swarm.ConnMgr || {}
-  config.Swarm.ConnMgr.GracePeriod = '1m'
+  config.Swarm.ConnMgr.GracePeriod = '3m'
   config.Swarm.ConnMgr.LowWater = 20
   config.Swarm.ConnMgr.HighWater = 40
```
SgtPooki, line 52: @lidel Are daemon values configurable in the app? If not, is there a reason other than dev bandwidth?

RubenKelevra: @SgtPooki yes. You can modify the whole config file of the daemon.

dev bandwidth?

lidel: Yes, we simply had no bandwidth to work on this + waiting for resource manager to ship in go-ipfs 0.13, as noted in https://github.com/ipfs/ipfs-desktop/pull/2096#pullrequestreview-936752564

If you want to pick this up, bump ipfs-desktop to 0.13-rc1 as part of this PR.

lidel: - Let's keep this PR to be about one thing: `Swarm.ConnMgr.GracePeriod`
- I don't understand how LAN dials could kill routers, afaik it is high churn of WAN connections that is  the usual problem, but  if you have additional information please comment in https://github.com/ipfs/go-ipfs/issues/3320 
- Pubsub is tracked in https://github.com/ipfs/go-ipfs/issues/6621 and  https://github.com/ipfs/go-ipfs/issues/8591 – no updates.
- Hole punching will be enabled  in https://github.com/ipfs/ipfs-desktop/issues/1946#issuecomment-1131511214, no need to do anything

SgtPooki (approved):

SgtPooki:
> I propose we park this until at least one of below is available and implemented in this PR
> 
> * [ ]  go-ipfs 0.13 ships `Swarm.ResourceMgr` – that will allow us to set hard connection limits
> * [ ]  we are able to set separate Grace Period  for peers that sent us some data over bitswap

@lidel are there tracking issues for these? Kubo is on 0.15 now, so I assume that first item is done. What is needed to handle that second item?

SgtPooki (approved):

SgtPooki:
Hey Ruben,

Thanks so much for this PR and for your patience with us. After talking with Lidel, we've agreed that the proper fix is to remember peers you connected to when viewing a DAG, and re-use them when attempting to revisit the DAG. This work would need to be done inside of Kubo. Would you mind opening an issue in the [Kubo](https://github.com/ipfs/kubo) repo for this feature enhancement?

If we were to increase the GracePeriod blindly for all connections, there are a large number of potential issues that could arise with this. We think this should be a user-based decision to increase this value.

PR #2125, opened 2022-05-04
1177 refactor ctx speed improvements

### LINKED ISSUE #1177: Refactor `ctx`
There is just one codebase that is run by one instance of the program at the time. Is it worth having a variable called `ctx` (context)? That'd certainly be more useful in HTTP (or any other protocol) handlers where each request is different.

While working on #1175 I found it a bit annoying to have to pass `ctx` around because I needed one single variable.

The idea is to remove `ctx` where possible and simply export the functions. Also, refactor the code so we don't have named and default exports on the same file as we do right now. That has been the cause for some bugs up until now.

SgtPooki:
an extension on https://github.com/ipfs/ipfs-desktop/pull/2114 which fixes #1177; this PR is the result of me investigating various problems with the load times and intermingled dependency graphs between src modules


SgtPooki:
superseded by https://github.com/ipfs/ipfs-desktop/pull/2378

PR #2140, opened 2022-05-13
chore: remove deprecated ipfs-on-path, npm-on-ipfs

### LINKED ISSUE #1935: From where it assumes the user home path?
/home/.ipfs-desktop/IPFS_PATH is assigning a wrong username every time I start IPFS Desktop

Probably this bug affects from version 13 (or bellow) to the latest 17

hacdias:
Back in February 2021, we deprecated `ipfs-on-path` (#1768) and in January 2022 we added some code to uninstall both features completely (#1948, #1947).

Since @SgtPooki is working on many maintenance improvements right now, I thought it'd be a great time to get rid of this chunk of code that is deprecated. It also decreases the scope of some other maintenance issues as there won't be a need to touch this code anymore.

Closes #1935


lidel (approved):
sgtm, thank you for cleaning this up!

PR #2152, opened 2022-05-20
Remove hardcoded DHT client routing

### LINKED ISSUE #1859: Ask user if he want's to enable dht (instead of just running a client)
Currently, the default config is using `Routing.Type = 'dhtclient'`. This isn't really scaling if we onboard more and more desktop clients.

We should ask the user if he/she if she likes to contribute more to the network and has a stable non-metered internet connection + runs the computer usually for several hours.

Something along these lines.

This would help the network in the long run, as the desktop users get aware of the fact that they do run an ipfs node, but only using network resources.

If a node isn't properly accessible autonat should disable the dht server functionality anyway, right?

2color:
Closes #1859

This PR removes the hard-coded dht client mode and instead relies on go-ipfs' defaults which test for the node's dialability and if so switch the DHT to server mode automatically.

A migration was added to remove the hard-coded routing configration from the users config.json.

### DIFF
--- a/src/common/store.js
+++ b/src/common/store.js
@@ -51,6 +51,16 @@ const migrations = {
       flags.push('--routing=dhtclient')
       store.set('ipfsConfig.flags', flags)
     }
+  },
+  '>=0.20.6': store => {
+    let flags = store.get('ipfsConfig.flags', [])
+
+    // use default instead of hard-coded dhtclient
+    const dhtClientFlag = '--routing=dhtclient'
+    if (flags.includes(dhtClientFlag)) {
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thank you, LGTM (my rationale in https://github.com/ipfs/ipfs-desktop/issues/1859#issuecomment-1131506875).

In general, we are moving away from CLI parameters, and use config these days.

If a user wants to change the default behavior, they can do that on _Settings_ screen, by explicitly setting [`Routing.Type`](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#routing) to `dhtclient`

PR #2159, opened 2022-05-26
feat: monochrome tray icon on Windows & Linux

hacdias:
This addresses #956 by adding monochrome tray icons for the remaining platforms (Windows and Linux). This way, we are consistent cross-platform and consistent with the platforms themselves.

**Background**, or, _why did we not do this before?_:

1. On Linux, I don't think we've ever had the conversation of supporting both light and dark themes. Therefore, it was just something we missed.
2. On Windows, there is a little caveat: Windows has both application and system themes and they are separate. Electron is only equipped to detect [application mode](https://github.com/electron/electron/issues/19729). However, recent versions of Windows (11+) do not separate both themes anymore! So I think it is safe to assume that Electron gives us the correct theme to use.
3. On macOS we don't really need different icons. We just give a "template" icon and macOS uses it as alpha channel, adapting to whatever theme the user has on their system ✨ 

Screenshots on Windows:

light-mode

dark-mode

@lidel @SgtPooki I would not consider this a priority. However, I would like to ask you to test this on any Linux platform you have in hand to check if it behaves correctly. If not, we can keep the colored icons for the time being.


lidel (dismissed):
>  On Linux, I don't think we've ever had the conversation of supporting both light and dark themes.   Therefore, it was just something we missed.

We did not touch this on Linux because (afaik) there is no standard API for this that works across Linux ecosystem.  Linux people use different DE and using the colored icon is safer.

As a data point, the  icon is invisible on old school black [i3bar](https://i3wm.org/i3bar/):

> 


@hacdias mind restoring old behavior (colored icon) on Linux?

SgtPooki:
> @hacdias mind restoring old behavior (colored icon) on Linux?

Instead, can we include a setting for this and default linux to the colored icon?

SgtPooki (changes requested):
Code LGTM but we should resolve the linux tray icon discussion

lidel:
I don't feel super strong about this, my default mindset is to avoid adding niche features that increase maintenance cost, but this is small enough that it should be fine.

SgtPooki (dismissed):

SgtPooki:
@hacdias can you confirm if you were planning on implementing the 

> setting for this[monochrome icon] and default linux to the colored icon

hacdias:
@SgtPooki I updated the PR such that a new option shows under **Settings** on Windows and Linux. By default, we keep the current behavior (colored icons), but users are free to choose if they want monochromatic icons or not.

Having a different default for Windows and Linux would involve further changes, to which I don't have bandwidth right now.

Tested on macOS (to check if the behavior is kept) and Windows (it works fine).

image

SgtPooki (approved):

PR #2207, opened 2022-07-18
chore: remove unused auto-project triggers

galargh:
By doing this we avoid starting the workflow in cases where we don't want to do any real work anyway.


SgtPooki (approved):

SgtPooki:
The workflow has been failing since I implemented it, but I agree with the changes in this PR.

I'm going to try to dive into that action to figure out whats going on, but we may need to write one ourselves in action-script.

PR #2208, opened 2022-07-18
chore: use official github action in auto-project

galargh:
Don't know if you saw this one - there's an action for adding to project which is maintained by GitHub - https://github.com/actions/add-to-project. In general, I think it's a safe default to use GitHub's things when they're good enough. Feel free to close if you want some of those extra features that the other action comes with.

disclaimer: not tested - just wanted to draw your attention to this


SgtPooki:
The other one is broken, so I think this would probably be the better call

SgtPooki (approved):

PR #2227, opened 2022-08-11
feat: jsdoc and improvements to daemon logic

hacdias:
While doing #2225, I noticed there are some few things could be improved:

- [x] Add JS Doc comments
- [x] Added custom error messages from things we already detected (config file missing, repository not being a directory) 
- [x] IPFS Desktop now quits if the user does not choose to change the port when they're busy. 

License: MIT

### DIFF
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -413,17 +413,19 @@ function checkRepositoryAndConfiguration (ipfsd) {
       return false
     }
 
-    if (!configExists(ipfsd)) {
-      // Config is generated automatically if it doesn't exist.
-      logger.error(`configuration does not exist at ${ipfsd.path}`)
-      dialogs.repositoryConfigurationIsMissingDialog(ipfsd.path)
-      return true
-    }
+    if (!apiFileExists(ipfsd)) {
+      if (!configExists(ipfsd)) {
+        // Config is generated automatically if it doesn't exist.
+        logger.error(`configuration does not exist at ${ipfsd.path}`)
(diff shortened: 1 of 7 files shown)


hacdias:
@SgtPooki could you take a look at this, please? I think it's a good start and there's a few places where I put `any` but some also depend on other parts of the code so I didn't add them yet.

src/daemon/config.js
```diff
@@ -238,19 +329,19 @@ async function checkPorts (ipfsd) {
...
+  const configApiMa = parseMultiaddr(config.Addresses.API)
+  const configGatewayMa = parseMultiaddr(config.Addresses.Gateway)
 
   const isApiMaDaemon = await checkIfAddrIsDaemon(configApiMa.nodeAddress())
   const isGatewayMaDaemon = await checkIfAddrIsDaemon(configGatewayMa.nodeAddress())
 
   if (isApiMaDaemon && isGatewayMaDaemon) {
     logger.info('[daemon] ports busy by a daemon')
-    return
+    return true
```
SgtPooki, line 340: > Returns true if ports are cleared for IPFS to start.

Do we want to return true if the ports are busy?

SgtPooki (changes requested):
A few minor questions and changes are requested.

src/daemon/config.js
hacdias: Ah, this is a specific case: if ports are busy by a **daemon**, it shouldn't be a problem since we connect to it anyways (via the API). The problem is when the ports are busy by something else.

SgtPooki (approved):

PR #2268, opened 2022-09-14
Use GitHub Dependabot to manage deps

VoidxHoshi:


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
SgtPooki, line 6: This is good, so we can keep our github-actions up to date.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
SgtPooki, line 11: Our backlog is already pretty difficult to manage, we would want to keep the noise down on updates here. We would want to limit the number of open PRs and also set up auto-merge on CI passing. Would you mind opening a PR to https://github.com/ipfs/github-mgmt so that we get this across all of our repos.

SgtPooki: https://github.com/ipfs/github-mgmt/tree/master/files/.github/workflows

PR #2344, opened 2022-11-22
chore: release prep

lidel:
- [x] kubo 0.17.0 (#2338)
- [x] bump some dependencies, if not a time sink
- [x] locale sync


galargh:
> Seems the new version of electron-builder no longer does signing on PRs, which in turn breaks notarization.
> 
> This seems to be a precaution for circleci, which we dont use – we use github actions, and afaik secrets are not exposed on PRs from forks (@galargh is that still true?).
> 
> I can either undo `electron-builder` update, or see if we can skip / enable signing on PRs without breaking the build.

Yes, that's correct. Workflows triggered on `pull_request` in general don't have access to secrets. `pull_request_target` does but we don't use it here.

lidel:
Added two secrets to enable signing when secrets are present.

https://github.com/ipfs/ipfs-desktop/blob/55f00514371decd853afe02aee03f5bde958bc68/.github/workflows/ci.yml#L216-L218

This restored macOS signing on non-fork PRs.

lidel:
@galargh I'm merging this to unblock you.

We want to make a new IPFS Desktop release to include Kubo 0.17 from #2338.

Release steps are in https://github.com/ipfs/ipfs-desktop/blob/main/DEVELOPER-NOTES.md, but feelf ree to reach out if you need any assistance.

PR #2352, opened 2022-12-02
docs(readme): add repology badge

lidel:
This PR adds a badge and note about third-party packages and the trade-off around autoupdate mechanism.

Using the badge from repology.org removes the burden of tracking/updating any of this:

[Packaging status](https://repology.org/project/ipfs-desktop/versions)


hacdias (approved):
Ameezing!

PR #2354, opened 2022-12-03
Create devcontainer.json

sawyeriii:
it


SgtPooki:
@sawyeriii Thanks so much for the contribution but I don't think this is a great time for us to merge this PR. 

For one, to add .devcontainer support for devs, we would need to update the contributors guide, or new markdown file linked from the contributors guide and/or FAQ. 

Two, we don't have time to prioritize ensuring this developer workflow, and documentation around it, work sufficiently. 

If you're experiencing issues setting up the project with the existing documentation, we would be happy to help you once we receive an issue targeting the specific problem you're running into.

PR #2363, opened 2022-12-15
feat: kubo 0.18 with /quic-v1 and /webtransport

galargh:
Related to https://github.com/ipfs/kubo/issues/9417

This PR is to upgrade Kubo to v0.18.0-rc1.

### DIFF
--- a/patches/multiaddr+10.0.1.patch
+++ b/patches/multiaddr+10.0.1.patch
@@ -1,0 +1,34 @@
+diff --git a/node_modules/multiaddr/src/convert.js b/node_modules/multiaddr/src/convert.js
+index c315201..80170c7 100644
+--- a/node_modules/multiaddr/src/convert.js
++++ b/node_modules/multiaddr/src/convert.js
+@@ -51,6 +51,7 @@ Convert.toString = function convertToString (proto, buf) {
+     case 55: // dns6
+     case 56: // dnsaddr
+     case 400: // unix
++    case 466: // certhash
+     case 777: // memory
+       return bytes2str(buf)
+ 
+@@ -84,6 +85,7 @@ Convert.toBytes = function convertToBytes (/** @type {string | number } */ proto
+     case 55: // dns6
+     case 56: // dnsaddr
+     case 400: // unix
++    case 466: // certhash
+     case 777: // memory
+       return str2bytes(str)
+ 
+diff --git a/node_modules/multiaddr/src/protocols-table.js b/node_modules/multiaddr/src/protocols-table.js
+index 3431af5..8939fb1 100644
+--- a/node_modules/multiaddr/src/protocols-table.js
++++ b/node_modules/multiaddr/src/protocols-table.js
+@@ -60,6 +60,9 @@ Protocols.table = [
+   [445, 296, 'onion3'],
+   [446, V, 'garlic64'],
+   [460, 0, 'quic'],
++  [461, 0, 'quic-v1'],
++  [465, 0, 'webtransport'],
++  [466, V, 'certhash'],
+   [477, 0, 'ws'],
+   [478, 0, 'wss'],
+   [479, 0, 'p2p-websocket-star'],
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "i18next-fs-backend": "1.1.4",
         "i18next-icu": "^2.0.3",
         "intl-messageformat": "^9.13.0",
-        "ipfs-http-client": "^60.0.0",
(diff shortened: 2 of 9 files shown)


galargh:
@lidel Do you have any ideas on what might be going on here https://github.com/ipfs/ipfs-desktop/actions/runs/3705807808/jobs/6280210555?

lidel:
Oof, yes, this is a fun one:  Kubo 0.18 is listening on `/quic-v1` multiaddrs by default.

IPFS DEsktop uses an old version of ipfsd-ctl that depends on old version of multiaddr without https://github.com/multiformats/js-multiaddr/pull/294, and that combination fails to start due to unknown protocol:

```console
$ npm start
2022-12-16T00:42:40.895Z error: [ipfsd] start daemon Error: no protocol with name: quic-v1
...
```

Upgrading to latest libs is soft-blocked until we finish https://github.com/ipfs/kubo/issues/9125 (subjective, but my gut feeling is it makes no sense to invest time in old library if we need to switch anyway).

For now, did the same thing we did in  https://github.com/ipfs/ipfs-webui/pull/2073 and patched-in new protocol support. This unblocks version bump to Kubo 0.18 :)

lidel:
Added config migrations to ensure Desktop and Brave (https://github.com/brave/brave-browser/issues/27965) will run with the same settings.

CI fails due to https://github.com/protocol/bifrost-infra/issues/2300, can work around this by re-running ot using `GO_IPFS_DIST_URL=https://dist-ipfs-tech.ipns.cf-ipfs.com` if needed.

galargh:
I updated my local node to 18.13 and modified the tests so that I can see what the actual errors are:
```
Error: TypeError: RequestInit: duplex option is required when sending a body.
```

which looks like https://github.com/nodejs/node/issues/46221. 

I suspect we're going to have to patch that in https://github.com/search?q=repo%3Aipfs%2Fjs-ipfs-utils%20fetch&type=code but I haven't pinned it down exactly just yet as I run out of time for now. I might try to give it another go in the evening.

galargh:
I think updating `ipfs-http-client` might help. I'll try it out later.

galargh:
https://github.com/ipfs/ipfs-desktop/pull/2363/commits/e6528d369b9dde8ca934266b5c7c78cb542cbef6 `Test` fixed. Now it's `Test end-to-end` that keeps acting up.

galargh:
I think it's this - https://github.com/ipfs/ipfs-desktop/blob/e6528d369b9dde8ca934266b5c7c78cb542cbef6/src/daemon/daemon.js#L34 - that's failing when running in electron.

galargh:
Going for a more minimal update and updating only [ipfs-utils to v9.10.0](https://github.com/ipfs/ipfs-desktop/pull/2363/commits/07ed2878ab55e7d68b5aa7c9eba79095aafa2b9c) instead of [ipfs-http-client to v60.0.0](https://github.com/ipfs/ipfs-desktop/pull/2363/commits/e6528d369b9dde8ca934266b5c7c78cb542cbef6) seem to have worked 🥳

patches/multiaddr+10.0.1.patch
```diff
@@ -0,0 +1,34 @@
...
+@@ -60,6 +60,9 @@ Protocols.table = [
+   [445, 296, 'onion3'],
+   [446, V, 'garlic64'],
+   [460, 0, 'quic'],
++  [461, 0, 'quic-v1'],
++  [465, 0, 'webtransport'],
++  [466, V, 'certhash'],
+   [477, 0, 'ws'],
+   [478, 0, 'wss'],
+   [479, 0, 'p2p-websocket-star'],
```
SgtPooki, line 34: I thought we updated multiaddr to 11? @whizzzkid @achingbrain

SgtPooki (approved):
questions about multiaddr changes that Alex and Nishant were doing, but I'm good with this otherwise.

patches/multiaddr+10.0.1.patch
achingbrain: This is all in `multiaddr@11.4.x` so you shouldn't need to monkey patch this.

lidel: IIUC we have no bandwidth to update things to latest versions, and that was blocking shipping Kubo release to Desktop users. Many users are in similar situation. 

Taking step back, Kubo update should not FORCE update of entire JS stack.
But the way things are today, `ipfs.swarm.peers` will crash when the backend suddenly gets updated to a version that returns new, unknown protocol. See https://github.com/ipfs/js-kubo-rpc-client/issues/103

Issue #2577, opened 2023-08-12
[IPFS Desktop] TypeError: Cannot destructure property 'stopIpfs' of 'undefined' as it is undefined.


ZeroCool22:
👉️ This error happens when i try to move the **Repository Location**:

https://github.com/ipfs/ipfs-desktop/assets/13344308/34fec5ce-f6d0-481c-a620-ee0f932c75c7

**Specifications**

- **OS**: Windows 10 Pro Versión 10.0.19045 Compilación 19045
- **IPFS Desktop Version**: 0.30.0
- **Electron Version**: I don't know?
- **Chrome Version**: It's IPFS Desktop

**Error**

```
Initializing daemon...
Kubo version: 0.21.0
Repo version: 14
System version: amd64/windows
Golang version: go1.19.10 


    at makeError (C:\Users\ZeroCool22\AppData\Local\Programs\IPFS Desktop\resources\app.asar\node_modules\execa\lib\error.js:60:11)
    at handlePromise (C:\Users\ZeroCool22\AppData\Local\Programs\IPFS Desktop\resources\app.asar\node_modules\execa\index.js:118:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
2023-08-12T05:11:09.756Z error: Cannot destructure property 'stopIpfs' of 'undefined' as it is undefined.
2023-08-12T05:11:38.169Z error: Cannot destructure property 'stopIpfs' of 'undefined' as it is undefined.
2023-08-12T05:15:09.295Z error: Cannot destructure property 'stopIpfs' of 'undefined' as it is undefined.
2023-08-12T05:16:52.074Z error: Cannot destructure property 'stopIpfs' of 'undefined' as it is undefined.
```

yatounoneko:
Have a same problem, also GUI is broken...

ZeroCool22:
> Have a same problem, also GUI is broken... I tried going back to 0.29.0 and everything works fine. ![image](https://user-images.githubusercontent.com/22129910/260215898-80aebae6-8e91-46f7-9f87-12f8e82c6f5c.png)

How i make a downgrade without losing anything?

yatounoneko:
> > Have a same problem, also GUI is broken... I tried going back to 0.29.0 and everything works fine. ![image](https://user-images.githubusercontent.com/22129910/260215898-80aebae6-8e91-46f7-9f87-12f8e82c6f5c.png)
> 
> How i make a downgrade without losing anything?

[...]

PR #2593, opened 2023-08-22
fix(tray): :adhesive_bandage: Fixes the context loaded async in the tray.

whizzzkid:
Fixes: #2577

In this PR:
- Passing the missing context in tray methods.
- This is a bandage fix to push out a working build.
- prolly release this as `v0.30.1`

TODO: Move this to better type-definitions to not run into this.

### DIFF
--- a/src/tray.js
+++ b/src/tray.js
@@ -159,7 +159,6 @@ async function buildMenu () {
         {
           id: 'runGarbageCollector',
           label: i18n.t('runGarbageCollector'),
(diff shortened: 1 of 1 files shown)


SgtPooki (changes requested):
i've got some changes local.. just gonna test now

SgtPooki (approved):
@whizzzkid take a peek at these new changes when you get a chance, and then i'm good with it?

PR #2634, opened 2023-09-20
docs(readme): add repo migration FAQ sections

SgtPooki:
Adds one new main section, with two subsections, to the https://github.com/ipfs/ipfs-desktop#faq--troubleshooting section of the README:

1. Error: Initializing daemon...
    * Error: Your programs version (N) is lower than your repos (N+x).
    * Found outdated fs-repo, migrations need to be run. - Error fetching: context deadline exceeded

Hopefully these instructions will help users unblock themselves, as well as give us give us a standard place to send users who run into these issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,81 @@ When upgrading, IPFS may need to perform migrations and for that we need a stabl
 2. Make sure your firewall or antivirus is not blocking requests, such as P2P traffic;
 3. Try again, by restarting IPFS Desktop.
 
+### Error: Initializing daemon...
+
+These errors pop up from [ipfsd-ctl](https://github.com/ipfs/js-ipfsd-ctl) when the Kubo daemon fails to start up. Below are some scenarios where you may run into this error.
+
+#### Error: Your programs version (N) is lower than your repos (N+x).
+
+This means you are attempting to run an older version of ipfs-desktop or Kubo than you have previously ran on your machine. Each Kubo version (which is included with ipfs-desktop) is tied to a specific IPFS repo version, which you can see at https://github.com/ipfs/fs-repo-migrations#when-should-i-migrate.
+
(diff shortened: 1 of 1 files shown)


SgtPooki:
also see https://github.com/ipfs/ipfs-desktop/issues/2635

README.md
```diff
@@ -257,6 +257,81 @@ When upgrading, IPFS may need to perform migrations and for that we need a stabl
...
+```
+
+##### Windows (powershell)
+
+```powershell
+cp -r ~/.ipfs ~/.ipfs.bak # backup current .ipfs folder. The path may be different, see [repository](#how-does-ipfs-desktop-select-the-ipfs-repo-location).
+cd C:\Users\myUserName\Downloads
+./fs-repo-migrations.exe -revert-ok -to N # Where N is the current repo version
+
+```
```
lidel, line 289: I think this is a power user option, last resort for data recovery. 
If someone wants to do this, they don't need this level of hand-holding.
Maybe trim to a single paragraph?

```suggestion
The ideal solution is to ensure you're running the latest version of ipfs-desktop, as upward migrations happen automatically. However, if you need to run the older version that is emitting this error, you will need to run a migration in reverse, manually. You can follow the official instructions [here](https://github.com/ipfs/fs-repo-migrations/blob/master/run.md) but with additional parameters: `fs-repo-migrations -revert-ok -to N`. See `fs-repo-migrations --help` for more information.
```

lidel: Ack, thanks! Btw: if you want this to be useful to non-power users, you should include information how to disable autoupdate, otherwise the fix will be reverted on reboot.

README.md
```diff
@@ -257,6 +257,60 @@ When upgrading, IPFS may need to perform migrations and for that we need a stabl
...
+
+For this method, you have to have the `ipfs` binary available on your command line:
+
+```bash
+ipfs config --json Migration.DownloadSources '["IPFS", "https://dweb.link", "https://cloudflare-ipfs.com", "HTTP"]'
+```
+
+##### Manually in an editor (not recommended)
+
+You can also edit the config file (`~/.ipfs/config` or `C:\Users\Username\.ipfs\config`) manually. Just make sure the json file is valid when you finish.
```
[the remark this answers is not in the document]
SgtPooki: That actually points to ipfs-desktop config files; though it should be updated to point to ipfs config/repo dir

PR #2651, opened 2023-10-05
fix: ipfs-desktop config can disable autoUpdate

### LINKED ISSUE #2613: How i can block IPFS Update?
I want to use the **0.29.0** version and not want it to get updated, what file i need to block in the firewall?

SgtPooki:
see https://github.com/ipfs/ipfs-desktop/issues/2613#issuecomment-1747792579 for more details

fixes #2613


SgtPooki:
I talked to lidel about it. I mean, our #1 fix suggestion is to upgrade to latest.. So if people do turn off autoUpdate we wont be able to do much.

But thats also why I dont want a UI option for it

src/auto-updater/index.js
```diff
@@ -7,8 +7,14 @@ const { showDialog } = require('../dialogs')
 const { IS_MAC, IS_WIN, IS_APPIMAGE } = require('../common/consts')
 const ipcMainEvents = require('../common/ipc-main-events')
 const getCtx = require('../context')
+const store = require('../common/store')
+const CONFIG_KEYS = require('../common/config-keys')
 
 function isAutoUpdateSupported () {
+  if (!store.get(CONFIG_KEYS.AUTO_UPDATE, true)) {
```
lidel, line 14: could we flip this and have `DISABLE_AUTO_UPDATE` instead? (we don't want to disable autoupdates due to a bug where there is a typo in key  or the key is missing (`!undefined` → `true` :grimacing: )

```suggestion
  if (store.get(CONFIG_KEYS.DISABLE_AUTO_UPDATE, false)) {
```

src/common/config-keys.js
```diff
@@ -6,6 +6,7 @@
 const CONFIG_KEYS = {
   AUTO_LAUNCH: 'autoLaunch',
   AUTO_GARBAGE_COLLECTOR: 'automaticGC',
+  AUTO_UPDATE: 'autoUpdate',
```
lidel, line 9:
```suggestion
  DISABLE_AUTO_UPDATE: 'disableAutoUpdate',
```

lidel (changes requested):
lgtm, but it should be a flag that controls disabling, not enabling  – see comment inline

lidel (approved):
(that is to say, ok to ship once feedback above addressed)

PR #2662, opened 2023-10-10
ci: pin node-version to 18.17.1

galargh:
Node 18.18.0 breaks `npm ci` command on Windows.


SgtPooki (approved):
Thanks @galargh I was diving down all the wrong holes here.

PR #2829, opened 2024-06-12
docs: add more known soutions

lidel:
I made a sprint over all open "gui error report" issues and tracked down ones that had a lot of duplicates. Same problems reported over and over, and same FAQ/solutions copied and pasted.

This PR aims to leverage duplicate error reports and existing solutions to point users who experience them towards an existing guide / FAQ, rather than having them open another duplicate.

- [x] go over user reports and identify duplicates
- [x] add link for each solution to the switch in `./src/dialogs/errors.js`
- [x] allow users to self-service by accessing kubo config when daemon failed to start


lidel:
Ok, landing this to include in v0.36  (https://github.com/ipfs/ipfs-desktop/pull/2828).

This way, the majority of recurring problems will be routed toward solutions, and we will spend less time on triage.

PR #2914, opened 2025-01-17
Create FUNDING.json

cewood:
Add FUNDING.json indicating that this repo is owned/maintained by Shipyard.


lidel (approved):
CI fails due to unrelated change at github, we will investigate and fix it in separate PR, when shipping kubo 0.33.

PR #2915, opened 2025-01-24
fix(ci): ubuntu e2e tests

lidel:
WIP, will be finding a way to fix regression identified  in #2914, likely due to upstream issues:
- https://github.com/electron/electron/issues/42510
- https://github.com/microsoft/playwright/issues/34251


lidel:
This should be enough until https://github.com/electron/electron/issues/42510 is resolved upstream in Electron.

PR #2918, opened 2025-01-30
fix: ipfs-webui v4.4.2

lidel:
https://github.com/ipfs/ipfs-webui/releases/tag/v4.4.2


lidel:
Note to self: CI fails at 
- https://github.com/ipfs/ipfs-desktop/actions/runs/13062008753/job/36447036602?pr=2918#step:7:214

because upload-artifact@v4  change, infor about potential fix at
- https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

PR #2957, opened 2025-06-10
integrated prettier

Pfed-prog:
Linted files with prettier

Currently `lint` produces such output
image


Pfed-prog:
> We do not want to use prettier. We will use standard as defined at https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#linting--code-style

There is no aegir in the repository

PR #2960, opened 2025-06-12
Integrated aegir as lint

Pfed-prog:


Pfed-prog:
Fixing error type by wrapping in String()
image

image

Pfed-prog:
removing enableRemoteModule from `BrowserWindow` Electron module

last mentions in the docs:
https://www.electronjs.org/blog/electron-10-0#breaking-changes

Why?
https://www.electronjs.org/docs/latest/breaking-changes#default-changed-enableremotemodule-defaults-to-false

> We [recommend moving away from the remote module](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31).

And figured out how to access a `height` and `width` property by following docs
https://www.electronjs.org/docs/latest/api/structures/display

Pfed-prog:
Further Updating `getUserMedia`
https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
outdated API for getting media streams from the user's device. The mandatory object and properties like chromeMediaSource are specific to older implementations and may not be compatible with modern browsers.

package.json
```diff
@@ -7,7 +7,7 @@
   "main": "src/index.js",
   "scripts": {
     "start": "cross-env NODE_ENV=development electron .",
-    "lint": "ts-standard && standard",
+    "lint": "aegir lint",
```
lidel, line 10: 💭 the fact we are adding ~30k of lines to the `package-lock.json` gives me a pause. it is pulling a lot of dev dependencies.

the only reason aegir is pulled is linter - it does not sound right. 
can the linter be updated without pulling in aergir?

lidel (changes requested):
@Pfed-prog thank you for opening this, but we can't accept this PR in the current form:

- these `getUserMedia` and `BrowserWindow`  fixes should be separate PRs. It is time-consuming to review when you have thousands of unrelated lines changed in this PR due to linter.
  - please remove these changes from this linter PR, and open a new PR per each of them.
- the linter fixes are fine, as long you remove dependency on `aegir` and only depend on linter. (adding  30k of lines to `package-lock.json` just for linter does not feel acceptable)

package.json
Pfed-prog: Thanks for providing feedback. I do really appreciate it. 

But I was advised to add/use Aegir. I think that the middle ground here is to remove aegir after we agree on the other commits.

richmondsasha0 (approved):

Pfed-prog:
Makes a lot of sense

I do believe that aegir related lint improvements could be discarded

But I have created smaller prs based on this large pr

PR #2966, opened 2025-07-13
MediaDevices Mozilla API outdated

Pfed-prog:
getUserMedia
https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
outdated API for getting media streams from the user's device. The mandatory object and properties like chromeMediaSource are specific to older implementations and may not be compatible with modern browsers.


lidel (changes requested):
Please avoid changing 49 files if it is not required to fix the problem. If you feel there is a need for fixing TS, make that in a separate PR.

Pfed-prog:
Sure, updated as asked in https://github.com/ipfs/ipfs-desktop/pull/2973

PR #2967, opened 2025-07-13
BrowserWindow Configuring correctly

Pfed-prog:
image

removing enableRemoteModule from BrowserWindow Electron module

last mentions in the docs:
https://www.electronjs.org/blog/electron-10-0#breaking-changes

Why?
https://www.electronjs.org/docs/latest/breaking-changes#default-changed-enableremotemodule-defaults-to-false

We [recommend moving away from the remote module](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31).

And figured out how to access a height and width property by following docs
https://www.electronjs.org/docs/latest/api/structures/display

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Use IPFS Desktop to get acquainted with IPFS without needing to touch the termin
 
 When in doubt, pick one of package formats with built-in automatic update mechanism:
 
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
Please avoid changing 49 files if it is not required to fix the problem. If you feel there is a need for fixing TS, make that in a separate PR.

Pfed-prog:
created https://github.com/ipfs/ipfs-desktop/pull/2974

PR #2970, opened 2025-07-13
Log errors

Pfed-prog:
Removed ts-ignore and fixed error log due to type error


lidel (changes requested):
@Pfed-prog thank you for submittin the fix, however this PR changes 49 files for no reason (if the purpose is to fix log error), thus making it difficult to review.

Are you able to make this PR only about the fix?

PR #2983, opened 2025-08-15
fix:macos ipfs-desktop left click no focus

SAHU-01:
This fixes https://github.com/ipfs/ipfs-desktop/issues/2943 

**_Problem:_** On macOS, left-clicks on WebUI inputs didn’t focus because they sat inside an Electron draggable region (app-region: drag) that swallowed pointer events.
**_Solution_**: Injected CSS from the Electron shell to mark interactive controls as app-region: no-drag, restoring normal left/right click behavior across loads and in-page navigation.

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -141,6 +141,24 @@ module.exports = async function () {
 
   openExternal()
   const window = createWindow()
+
+if (process.platform === 'darwin') {
+  window.webContents.on('did-finish-load', () => {
+    window.webContents.insertCSS(`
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
Thank you for fixing this @SAHU-01 ❤️ 

I think we want this to be fixed for all webui users, not just desktop ones. 
Would it be ok to close this and instead fix this upstream in webui in https://github.com/ipfs/ipfs-webui/pull/2416?

PR #3033, opened 2025-12-17
refactor: update code for ESM compatibility

### LINKED ISSUE #3025: replace proxyquire in tests
[`proxyquire`](https://www.npmjs.com/package/proxyquire) hasn't had a release on NPM since 2019 and has no support for modern JavaScript modules/ESM https://github.com/thlorenz/proxyquire/issues/217#issuecomment-1143690177.

Node.js has this built-in https://nodejs.org/api/test.html#mockmodulespecifier-options

dynst:
With ESM, all exported values need to have a name,

- Closes #3025


lidel (changes requested):
@dynst thank you for opening this. unfortunately something broke in this PR  and tests no longer pass -- would you have bandwidth to investigate?

if this turns out a laborious effort, we may need to go with more surgical ESM support.

fwiw i had attempted "esm loader" draft in (wip) https://github.com/ipfs/ipfs-desktop/pull/3034 to avoid full ESM switch and changing too much, but it does not seem to be working (dynamic imports dont work in main electron thread)

PR #3035, opened 2025-12-18
fix: prevent port replacement from corrupting IP addresses

lidel:
the naive string replacement in checkPorts() could corrupt multiaddr IP addresses when the port number appeared in the IP portion.

for example, replacing port `0` in `/ip4/127.0.0.1/tcp/0` would match the first `0` in `127.0.0.1`, producing `/ip4/127.5001.0.1/tcp/0`.

this caused "invalid ip address" errors on startup when ports were busy.

fix by using `/tcp/${port}` pattern to match only the port component.

ref: https://discuss.ipfs.tech/t/invalid-ip-address/19925

### DIFF
--- a/src/daemon/config.js
+++ b/src/daemon/config.js
@@ -419,11 +419,11 @@ async function checkPorts (ipfsd) {
   }
 
   if (busyApiPort) {
-    config.Addresses.API = config.Addresses.API.replace(apiPort.toString(), freeApiPort.toString())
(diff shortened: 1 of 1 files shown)


lidel:
Windows binary for testing:
https://ipfs.io/ipfs/bafybeiclfrdzm7qq5ilczlcyf7gvovjibgtsyzppv64tbekwllp2zvfxme/ipfs-desktop-setup-0.47.0-win-x64.exe

PR #3048, opened 2026-02-09
Prep for electron-serve update - load ESM via dynamic import and disable v8-compile-cache

RubenKelevra:
Hey @lidel, first contribution from me here in a long while :)

This PR is prep work for updating `electron-serve`.

Newer versions are ESM-only, so `require('electron-serve')` breaks. To unblock that dependency update, this switches `electron-serve` initialization to a dynamic `import()` and awaits it during WebUI init so the `webui://`-scheme is registered before any WebUI URLs are loaded. It also disables v8-compile-cache, which was causing `import()` to fail in this process.

Testing: I can only test on Linux. I've tested core functionality of the app under Linux with Electron `39.5.1` (start/stop daemon, WebUI navigation, tray menu actions, Files tab basics).

Outlook: I have more dependency updates and cleanups lined up. I'll send them as small consecutive PRs to keep each review focused and easy to approve.

### DIFF
--- a/src/webui/index.js
+++ b/src/webui/index.js
@@ -2,7 +2,6 @@
 const { screen, BrowserWindow, ipcMain, app, session } = require('electron')
 const { join } = require('path')
 const { URL } = require('url')
-const serve = require('electron-serve')
 const i18n = require('i18next')
 const openExternal = require('./open-external')
 const logger = require('../common/logger')
@@ -20,7 +19,16 @@ const ipcMainEvents = require('../common/ipc-main-events')
 const getCtx = require('../context')
 const { STATUS } = require('../daemon/consts')
 
-serve({ scheme: 'webui', directory: join(__dirname, '../../assets/webui') })
+const serveReady = Promise.resolve()
+  .then(() => import('electron-serve'))
+  .then((serveModule) => {
+    const serve = serveModule.default ?? serveModule
+    return serve({ scheme: 'webui', directory: join(__dirname, '../../assets/webui') })
+  })
+  .catch((err) => {
+    logger.error('[webui] failed to init electron-serve', err)
+    throw err
+  })
 
 /**
  *
@@ -123,6 +131,7 @@ const createWindow = () => {
 
 module.exports = async function () {
   logger.info('[webui] init...')
+  await serveReady
   const ctx = getCtx()
 
   if (store.get(CONFIG_KEY, null) === null) {


lidel:
Hey @RubenKelevra, thanks for all this work.
I am still busy with kubo 0.40 work, but by looking at the number of opened PRs you've spent a lot of time modernizing things..

Before I start reviewing, I want to raise a process concern. Right now there are 64 open PRs (#3048 through #3113) where each one carries all commits from every prior PR. PR #3113 has nearly 20k line change, 66 commits, 65 of which are duplicated from earlier PRs.

This makes review and testing impractical:

- For historical reasons the majority of  codebase is not covered by tests, thus each PR needs manual testing, but since they all bundle prior changes, I'd be re-testing nearly identical states 64 times.
- GitHub diffs show all accumulated changes, not just what's new in a given PR. You could switch target branch to previous PR, but that creates absurd chain of 60+ branches. 
- Any feedback on an early PR means rebasing all downstream PRs and re-running CI on each, which is expensive and takes away CI workers from other projects.
- Merging must happen in strict ascending order, one at a time.

The general rule I'd suggest: if a PR needs to include commits from other PRs to work, those changes all belong in one PR. Every PR from #3049 onward includes all commits from prior PRs, which means they're all effectively one unit of work. 

Could you consolidate into a single PR with all changes as individual commits? That way:

- I can still read the extra context from commit message and review commit-by-commit to understand each logical change if needed.
- I test the final state once instead of 64 times.
- If I have feedback on a specific change, it's just a fixup commit on a single branch -- not a cascade of rebases or cherry-picks across 60+ branches.
- The PR description can serve as a table of contents, grouping commits by area (deps, daemon, WebUI, etc.).

If any changes are truly standalone (no shared commits with the rest), those can stay as separate PRs WITHOUT "foreign" commits . But for everything that depends on other PRs in this series, a single PR would be much easier to work with.

I'd appreciate the consolidation -- the current state with 64 interdependent PRs is not something I'm able to review. If we had a single modernization PR that was a draft until you were done with modernization work, that would make review easier to plan.

 Hope you understand and once again thanks for all effort here.

lidel:
It is hard to answer your question about linter because I don't know which one of >60 PRs you mean :))

Without going into specifics, I know rules in this repoi are likely half a decade old, so likely ok to update them. 

--- 

**I'm going to close your PRs for now.** Right now they are 2 screens of PR noise which impacts our daily triage. 


@RubenKelevra  Please create a single one with everything, and keep it as draft until you feel its ready for review. Thanks again for your work, it just is not possible to review in current state.

RubenKelevra:
Hey @lidel,

Don't waste your time. I'll close / update the PRs today/tomorrow to make them reviewable. :)

PR #3122, opened 2026-02-27
fix(webui): navigation fixes

RubenKelevra:
This stack of changes is thematically all navigation fixes (from the previous PRs combined).

This fixes:
- First click navigation not working, and instead refreshing the status page
- Navigation in the UI -> closing window -> selecting tray navigation item opening the UI, but not navigating
- Application opening on a blank status page
- Navigation with tray menu navigation items sometimes not working properly after UI navigation
- ipfsd reload messing up the UI (can't remember what was happening in this case)
- duplicate url loads on duplicate selections of the same item
- Focus not properly set to the UI on restore
- `-3` errors in the log
- Broken error logging if value is not an error object (are now logged as "unknown error")

## Testing

I can only test on Linux, and I tested the core functionality of the app with Electron `39.5.1`.


lidel (approved):
Approach checks out. Thank you for fixing this annoying "initial state"  bug where one had to click twice to load Status 👍 

Pushed two commits on the branch:

- **ee80bd4 fix(webui): blank also on minimize and occlusion**
The new visibility flow only listens to `BrowserWindow` `'show'`/`'hide'`, which don't fire on user minimize, OS-level occlusion, or macOS app-hide (Cmd+H) / Mission Control, so the WebUI keeps re-rendering when the user can't see it. Restored the renderer-side `document.visibilitychange` listener alongside the IPC one (both feed the same idempotent updater) and added `'minimize'`/`'restore'` handlers in main. Comments call out that the goal is to avoid re-rendering the bandwidth graph on Status and the peer list on Peers so hopefully future LLMs won't miss this.

- **af50a78 fix(daemon): detect fatal kubo errors in listener**
Bundled to address the macOS upgrade-dialog e2e flake (reproduces on `main`). Root cause: `ipfsd.start()` rejects too slowly for the test's 45s budget, so the prompt stays on the in-progress template. Added `FATAL_KUBO_LOG_PATTERNS` so the listener flips `isErrored` directly from the version-mismatch line and the upgrade template loads without waiting.

PR #3123, opened 2026-02-27
feat(linux/autostart): include app path when launched via external electron

RubenKelevra:
Linux autostart currently writes only `process.execPath` into the Exec line. This works fine if Electron is bundled. However, I recently [switched to using the system's Electron-version](https://aur.archlinux.org/cgit/aur.git/commit/?h=ipfs-desktop&id=850dc8f822f455e4927459257af06fb54935f658) in my AUR-package. This requires to also supply the parameter to the `app.asar`, otherwise just electron39 gets started.

This upstreams [this patch](https://aur.archlinux.org/cgit/aur.git/commit/?h=ipfs-desktop&id=1778e3debb6e6725edbb422f1ed89c5bc0fe7c5f), to avoid patching locally.

It should not modify the current bundled app autostart creation.


lidel (approved):
LGTM. Pushed 8201ec7 to extend escaping to `$` and backtick per the [Desktop Entry Exec= spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html), and add unit tests covering the AUR (system electron +   `app.asar`) and bundled (electron-builder) launch layouts.

PR #3126, opened 2026-02-27
fix tray: use scale-aware lookup for status menu icons

RubenKelevra:
Status menu icons have used scale-suffixed filenames (`@1x`, `@1.5x`, `@2x`, ...) since the tray/menubar refactor in 2019, but the tray menu code still looked up a single `${color}.png` path.

That mismatch can break tray menu rebuilds at runtime when `${color}.png` is not present.

This change updates the lookup to pick the closest available scale variant for the current display scale factor, while still accepting `${color}.png` if it exists.

Testing: I can only test on Linux, and I tested the core functionality of the app with Electron `39.5.1`.


lidel (approved):
Thanks for catching this @RubenKelevra! I pushed a few follow-up commits to land a slightly different approach:

  - Use Electron's built-in HiDPI resolution (https://www.electronjs.org/docs/latest/api/native-image) instead of picking scales by hand: shipping an unsuffixed ${color}.png base alongside the @Nx variants is all that's needed. Covers every scale Electron supports (including `@1.25x`, the common Windows laptop default) and tracks the per-display scale on multi-monitor setups.
  - Extended gen_status.sh to emit the full DPI ladder plus the unsuffixed base, and ran optipng -o7 on the regenerated PNGs for a small size win.
  - Side-fixes so Windows CI actually runs: skip Azure code-signing on fork PR builds (secrets aren't available there), plus an afterEach teardown flake fix after merging main.

PR #3167, opened 2026-06-08
feat: kubo 0.42, electron 42

lidel:
- [x] https://github.com/ipfs/kubo/issues/11227
- [x] https://github.com/ipfs/ipfs-desktop/blob/main/DEVELOPER-NOTES.md#manual-testing
  - [x] windows
  - [x] linux
  - [x] macos


lidel:
@SocketSecurity ignore npm/js-yaml@4.2.0
@SocketSecurity ignore npm/string.prototype.padend@3.1.1

Socket comments are false positive. Both alerts are Socket's `obfuscatedFile` heuristic misfiring on minified dist bundles:

- `js-yaml@4.2.0` is the canonical YAML parser maintained by nodeca, pulled in transitively by `electron-builder`.
- `string.prototype.padend@3.1.1` is the standard es-shim polyfill by ljharb, pulled in transitively by `npm-run-all`.

Both are well-known, widely-used packages and arrive only as dev/build dependencies, not runtime code shipped to users.

PR #3170, opened 2026-06-15
fix(updater): mute transient update-check errors

mikelsr:
I leave my MacOS workstation on most of the time and put it to sleep when I'm not using it.
I have auto update enabled for IPFS Desktop and do my best to keep it updated.

That said, I get very frequent dialog panels with `Could not install the update automatically. Would you like to download it manually?` despite my installed version already being the latest.

I marked updates as "started" so if the update process couldn't even start the dialog is muted. I'm not sure this is the best approach, so any feedback from someone more familiar with the project is welcome. Thanks!

### DIFF
--- a/src/auto-updater/index.js
+++ b/src/auto-updater/index.js
@@ -22,6 +22,7 @@ function isAutoUpdateSupported () {
 
 let updateNotification = null // must be a global to avoid gc
 let feedback = false
+let updateStarted = false // set once the update has started
 
 function setup () {
   const ctx = getCtx()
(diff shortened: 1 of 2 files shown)


lidel (approved):
Thank you for the PR!
Your approach is right: only show the error on a manual check or during a download.

I've found and fixed one gap: `update-available` sets `updateStarted`, but only the `error` handler clears it, never `update-downloaded`. So after the first background download the flag stays `true`, and later transient errors pop the dialog again, the case you hit. Pushed e218bdf to clear it on `update-downloaded` (like on error), plus a regression test in 4e1ba46.

Let's wait for CI to finish, if green ok to merge.

PR #3186, opened 2026-07-16
fix(ci): skip windows signing without secrets

lidel:
## Problem

Since 10 June, every Dependabot PR's Windows build hangs for 40 minutes and then gets cancelled.

The build tries to sign the .exe, but Dependabot PRs cannot read our signing secrets. The credentials arrive empty, so the Azure signing tool opens a browser and waits for someone to log in. Nobody can log in on a CI runner. It used to fail fast instead, until electron-builder 26.15.2 dropped the check that caught this (electron-userland/electron-builder#9687).

We already skip signing when secrets are missing, but that step only looks for forks, and Dependabot branches live in this repo.

## Fix

- Check whether the Azure credential is really there, instead of guessing from the trigger type. One path covers forks, Dependabot, and anything else without secrets.
- If it is missing, drop the signing config so Windows still builds and uploads the .exe.

Signing on main is unchanged, and macOS already handled this on its own. These PR binaries are unsigned: fine for testing, not for shipping.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,20 +225,40 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         continue-on-error: true # release-please github-release will fail when a release-please PR was merged and the tag doesn't exist.
 
-      # PR builds from forks cannot access signing secrets. Strip Windows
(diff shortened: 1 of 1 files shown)


lidel:
Will test in https://github.com/ipfs/ipfs-desktop/pull/3177

PR #3210, opened 2026-09-06
fix: avoid duplicate daemon start during startup

RubenKelevra:
During packaging I noticed the following issue:

`setupDaemon()` currently starts Kubo and installs the `IPFS_CONFIG_CHANGED` restart listener in parallel with `setupAutoGc()`, `setupPubsub()` and `setupNamesysPubsub()`.

Those initializers can normalize the stored daemon flags. For example, `setupAutoGc()` adds `--enable-gc` when it is missing and emits `IPFS_CONFIG_CHANGED`.

If that happens while the initial daemon start is still running, the restart handler starts a second Kubo against the same repo. One of them then fails with:

`Error: lock .../repo.lock: someone else has the lock`

Run the three daemon flag initializers before `setupDaemon()`. They can still update the stored flags, but there is no restart listener yet and the daemon starts once afterwards with the final flags.

I've added a regression test which blocks all three initializers and verifies that `setupDaemon()` does not run until they finished.

### DIFF
--- a/src/daemon/index.js
+++ b/src/daemon/index.js
@@ -37,7 +37,19 @@ async function setupDaemon () {
     return status
   }
 
-  const startIpfs = async () => {
+  // Daemon transitions run one at a time. start and stop read `ipfsd` to
+  // decide what to do, and it only tells the truth between transitions:
+  // mid-start it is still null, so a second start would spawn a Kubo that
+  // races the first for the repo lock.
+  let transition = Promise.resolve()
+  const serialize = (fn) => () => {
+    const result = transition.then(fn)
+    // a failed transition must not reject or stall the ones queued behind it
+    transition = result.catch(() => {})
+    return result
+  }
+
+  const start = async () => {
     if (ipfsd) {
       return
     }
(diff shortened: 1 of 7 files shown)


RubenKelevra:
Windows test failure seems unrelated to my change, as it fails to find an ipfs binary for whatever reason as then just runs in a timeout.

Test seems to be overall flaky.

lidel (approved):
Thank you, LGTM. 

Pushed three commits on top:

- 02435fb **fix: keep splash first while gating daemon start**
  The barrier gated the whole wave, so the splash window was not built until all three initializers resolved.

- 489427b **fix: do not stall startup when the splash fails**
  Semi-related: `createSplashScreen` left the `splashScreen` context prop unset on failure, and nothing else resolves that deferred, so `setupWebUI` blocked on it forever.

- 1af7d82 **fix: run daemon transitions one at a time**
  `start` and `stop` read `ipfsd`, which is only accurate between transitions, so a tray toggle or an online-status flip landing mid-restart could still spawn a second Kubo.

PR #3211, opened 2026-09-06
fix: prefer distro launcher for autostart

RubenKelevra:
While packaging IPFS Desktop for Arch, I noticed there is currently no way for distro packages to have the autostart function use a custom path for setting it as startup.

This PR adds `IPFS_DESKTOP_EXEC` as an optional override. If it is set, the generated desktop entry uses that launcher instead of reconstructing the command from Electron and the app path.

This allows distro packages to have their wrapper scripts placed in the autostart by ipfs-desktop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,10 @@ Instead, update cycle is maintained by respective communities:
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/ipfs-desktop.svg)](https://repology.org/project/ipfs-desktop/versions)
 
+Packages that start IPFS Desktop through a wrapper script can set the `IPFS_DESKTOP_EXEC` environment variable to the path of that wrapper.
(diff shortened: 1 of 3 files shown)


lidel (approved):
Makes sense for distro packaging, thank you. Pushed two commits on top:

- 9c6ac39 **docs: document IPFS_DESKTOP_EXEC override**
  A README note so packagers find the variable, a comment on why the value is a path and not a command line, and a log line with the autostart file path and its `Exec=`, so a wrong launcher shows up in the log.

- 3987e4c **test: keep autostart cases free of ambient env**
  The `getLinuxAutostartExec` cases failed on any machine with `IPFS_DESKTOP_EXEC` exported, likely wasting someone's time to debug, or false-positive bug report. `withEnv` now sets or clears it like it already does with `execPath`.

Good to merge once CI is green.

